PR #19, opened 2015-12-16
Fix #18: Use EntityException, instead of Exception, through more of t…

### LINKED ISSUE #18: PassthroughMessageCodec exception use is asymmetric
PassthroughMessageCodec has 2 methods which are inverses, yet with asymmetric type uses:
  public static byte[] serializeObjectToArray(Exception exception)
  public static EntityException deserializeExceptionFromArray(byte[] bytes)

Ideally,
  public static byte[] serializeObjectToArray(Exception exception)
would become
  public static byte[] serializeExceptionToArray(EntityException exception)

That way, the name is more descriptive and the exception type is consistent.

jd0-sag:
…he internals and codec

-this starts with the change to make PassthroughMessageCodec's exception serialization symmetric but it allowed for a tightening of exception use-cases, elsewhere
-of key interest is the change to PassthroughServerMessageDecoder.MessageHandler's method signatures to throw EntityException, instead of Exception.  This means that more specific exceptions must be thrown from within the implementation, at the point where that knowledge is available.  The caller then catches RuntimeException to wrap anything else that slips through the cracks in an EntityUserException
-for the most part, this meant throwing EntityNotFoundException, instead of Exception, when an entity failed to be resolved

### DIFF
--- a/entity-test-lib/src/main/java/org/terracotta/passthrough/IFetchResult.java
+++ b/entity-test-lib/src/main/java/org/terracotta/passthrough/IFetchResult.java
@@ -18,11 +18,13 @@
  */
 package org.terracotta.passthrough;
 
+import org.terracotta.exception.EntityException;
+
 
 /**
  * Defines the callback invoked on successful resolution of a fetch message.  The reason why we need a callback and not a
  * simple call-return structure is that the fetch may block on lock acquisition.
  */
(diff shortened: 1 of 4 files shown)


ljacomet:
@jd0-sag Could you rebase this PR?

Issue #9, opened 2015-12-14
taracotta-parent inaccessible


myronkscott:
dependency on terracotta-parent 5.0.0.alpha1 dependency is inaccessible.  A new version of [taracotta-parent](https://github.com/Terracotta-OSS/terracotta-parent) needs to be published and the dependency should be updated here.

Once this dependency is fixed.  This repository should be built on cloudbees and published publicly

myronkscott:
linked to https://github.com/Terracotta-OSS/terracotta-parent/issues/5

PR #20, opened 2015-12-17
Issue #9

ljacomet:
Add repository information to enable community builds

### DIFF
--- a/entity-server-api/pom.xml
+++ b/entity-server-api/pom.xml
@@ -30,11 +30,4 @@
 
   <groupId>org.terracotta</groupId>
   <artifactId>entity-server-api</artifactId>
-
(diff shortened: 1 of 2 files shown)


myronkscott:
What about the version of terracotta-parent?  Should that change to 5.0-SNAPSHOT?

ljacomet:
I'd rather stick with released version. As indicated in terracotta-oss/terracotta-parent#5, there will be an upcoming 5.0.0.alpha2 that I will include in this PR.

ljacomet:
test this please

jd0-sag:
This exposed an indirect dependency on Guava, in a downstream project.  That has now been made direct so this can merge.

PR #31, opened 2016-01-14
Fix #30: Add IMonitoringProducer interface to standard-cluster-services

### LINKED ISSUE #30: Add a service interface for monitoring data
Entities and the platform need to provide various kinds of data for external monitoring tools.

Over time, this will grow and evolve to support various kinds of statistics and events but a starting point is the ability to provide structured data, describing the high-level state of the system.

This data will need to be largely open-ended in terms of what can be provided but the underlying implementation of the service receiving this data must remain very constrained to keep it simple are generally applicable.  This means that the service must receive the data, even though it can't process it.  Higher-level tooling will need to be built, external to the service, which can process and interpret the meaning of this data, as it will be specialized to which component produced it.

Additionally, this information must be complete, as it describes a state snapshot at any point in time and can't rely on an external tool always able to consume and interpret the data.  Hence, it can't be as simple as an event stream, but must be at least structurally meaningful to the service.

To solve this, the data will be structured as a tree.  The service will be able to understand how to add and remove nodes in the tree, but not interpret the data held in the leaves.

jd0-sag:
-this represents the starting-point of the service interface which can absorb arbitrary add/remove data element events from any entity or the platform

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/monitoring/IMonitoringProducer.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/monitoring/IMonitoringProducer.java
@@ -1,0 +1,31 @@
+package org.terracotta.monitoring;
+
+import com.tc.classloader.CommonComponent;
+
+
+/**
+ * The interface exposed by a monitoring service, allowing entities to push statistics and structured data which can be
+ * consumed by an external component (attached to the other side of the service, using some more specialized interface).
+ */
+@CommonComponent
+public interface IMonitoringProducer {
+  /**
+   * Adds a node to the internal data registry tree or replaces an existing one.
+   * By default, new nodes have no children.
+   * 
+   * @param parents The parent node names, starting from the root.
+   * @param name The name of the node to create or replace.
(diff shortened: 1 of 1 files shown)


standard-cluster-services/src/main/java/org/terracotta/monitoring/IMonitoringProducer.java
```diff
@@ -0,0 +1,31 @@
+package org.terracotta.monitoring;
+
+import com.tc.classloader.CommonComponent;
+
+
```
myronkscott: When we add the statistics based stuff (ring buffer) will that have a completely separate interface?  If so we might want to use a more specific name for this interface

jd0-sag: I assume that we will just augment this interface.

It will really depend on what that looks like and where we want to go with it.  I would like to avoid having to hold multiple resolved service interfaces in the consumer if we expect that both stats and events would be desired.

PR #38, opened 2016-02-04
Fix #27: Added tryDestroy to EntityRef API

### LINKED ISSUE #27: EntityRef could benefit from a non-blocking way of performing an entity destroy
Right now there is no way to attempt to destroy an entity and detect failure due to the holding of existing entity references.  This would seem like a useful addition to the api.

jd0-sag:
-this method is like destroy() but it will fail, instead of blocking, if it can't get the exclusive lock on the entity
-most of the change is updating passthrough to support this


jd0-sag:
Let me know if this API is what you wanted and I will either either re-assign it to Myron to merge or we will need to coordinate the merges (since there are some downstream repositories which depend on this).

chrisdennis:
I'm pretty sure this should work for me.  Lets go ahead and merge.

PR #40, opened 2016-02-05
Fix #35: Created MessageCodecException as checked exception for MessageCodec methods

### LINKED ISSUE #35: MessageCodec marshall/unmarshall methods should throw a checked exception
It would seem to make sense for a MessageCodec implementation to required to throw a specific checked exception type when presented with incorrect input.  What happens at the moment if I throw a runtime exception from a deserialize/serialize method?

jd0-sag:
-this gives the MessageCodec implementation a well-defined failure option

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/MessageCodec.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/MessageCodec.java
@@ -37,8 +37,9 @@ public interface MessageCodec<M extends EntityMessage, R extends EntityResponse>
    * 
    * @param payload The byte array containing the raw wire message
    * @return A high-level message instance
+   * @throws MessageCodecException The message could not be deserialized or was considered invalid.
    */
-  M deserialize(byte[] payload);
+  M deserialize(byte[] payload) throws MessageCodecException;
 
   /**
    * Deserializes a given message payload into a high-level message type, useful to the server-side entity.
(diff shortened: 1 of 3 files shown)


jd0-sag:
Chris:  Let me know if this is what you expected and had in mind.  I will get Myron to merge across the repos if you like this.

chrisdennis:
So on the client side I see this as a MessageCodecException wrapped inside an EntityException?  Should that be the case, or should the MessageCodecException be thrown in it's own right... is a failure in the codec a failure of the entity?  I guess this is back to the old question of how independent the codec is from the entity.

jd0-sag:
Yes, this approach will always wrap it in an EntityUserException instance.  The idea behind this being that it is a special kind of "user code exception" from the perspective of the platform.

Let me know if you think that something more specialized is needed but I view this as more of just a well-defined error case as opposed to a way to return highly specialized information.  I am not sure if that is the correct view, however.

chrisdennis:
From my point of view I just figure a codec failure would want to be handled differently to an actual entity failure... and that in the current scheme to do so I will need to know to unwrap the EntityUserException in to those two distinct cases.  That said this is moving in the same direction so we could go ahead with this merge, and reassess when I see how things pan out for real.  Make sense?

jd0-sag:
That makes sense.  We will proceed with this but may re-assess, in the future, if there is a meaningful difference between these cases from the perspective of the entity implementations.

That would probably just involve creating another sibling of EntityUserException, within the EntityException hierarchy, in which case we might want to make these methods just throw any Exception and we would wrap it (the entity can't generate the EntityException, itself, since it doesn't have visibility into the identifier the platform adds).

PR #48, opened 2016-02-17
Api refactoring

myronkscott:
Three commits.  

Move codec accessor from active entity to service and change generics type signature.
Move concurrency strategy to service from active entity
Add the ability to reconfigure an entity

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
@@ -45,13 +45,6 @@ public interface ActiveServerEntity<M extends EntityMessage, R extends EntityRes
    * @param clientDescriptor client-side instance which disconnected
    */
   void disconnected(ClientDescriptor clientDescriptor);
-
-  /**
-   * Get configuration for given entity
-   *
-   * @return serialized byte array
-   */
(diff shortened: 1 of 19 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -45,13 +39,6 @@
    * @param clientDescriptor client-side instance which disconnected
    */
   void disconnected(ClientDescriptor clientDescriptor);
-
-  /**
-   * Get configuration for given entity
-   *
-   * @return serialized byte array
-   */
-  byte[] getConfig();
```
jd0-sag: Are we ok with completely removing this method from the API, in this change?

jd0-sag:
My main questions with this revolve around why we are making all these changes to how config is handled:  one change introducing the ability to change it, after creation, while the other removes the ability to query it.  I would have thought these were interesting enough changes to be discussed, in isolation.  Reconfiguration might be hard to get right, as well, but that is arguable an implementation problem, not an API problem.

The size of the commits worries me, a little, but should be fine since they are already created (not only does this make the ideas in the change less clear but also makes me hope that we never need to do a bisect over these changes - that would be not very helpful).

We should talk through this, when you are free, later.

PR #50, opened 2016-02-23
Track transaction history of create/destroy/reconfigure for re-send short-circuit

jd0-sag:
-this removes the problematic shouldTolerateCreateDestroyDuplication() work-around to avoid errors in re-sent life-cycle messages
-we now track create/destroy/reconfigure messages so that consistent answers can be give, on re-send


jd0-sag:
The concern I voiced in tonight's call was a false-alarm.  The uses of "didAlreadyHandle", in the last commit, handle the case I think I missed in the tc-core implementation of the same functionality where it looks like we might still run the invoke, even if it is already handled.

This implementation doesn't have this problem so it should be good to merge whenever you are ready.

PR #51, opened 2016-02-28
fixes to support reconfigure on clustered map

myronkscott:
Some fixes to support reconfigure in passthrough.  

* do not take an exclusive lock for reconfigure.  It doesn't need it, MANAGEMENT_KEY exclusive mode provides enough isolation

* Services need to be completely rebuilt on restarts

*  close hook may not need to run after failover

### DIFF
--- a/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughEntityRef.java
+++ b/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughEntityRef.java
@@ -119,7 +119,6 @@ public class PassthroughEntityRef<T extends Entity, C> implements EntityRef<T, C
   public C reconfigure(C configuration) throws EntityException {
     // Make sure that we have a service provider.
     if (null != this.service) {
-      getWriteLock();
       try {
         byte[] serializedConfiguration = this.service.serializeConfiguration(configuration);
         PassthroughMessage reconfig = PassthroughMessageCodec.createReconfigureMessage(this.clazz.getCanonicalName(), this.name, this.version, serializedConfiguration);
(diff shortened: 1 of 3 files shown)


entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughServer.java
```diff
@@ -84,8 +84,9 @@ public synchronized Connection connectNewClient() {
       public void run() {
         synchronized (PassthroughServer.this) {
           PassthroughConnection closedConnection = PassthroughServer.this.savedClientConnections.remove(thisConnectionID);
-          Assert.assertNotNull(closedConnection);
-          PassthroughServer.this.serverProcess.disconnectConnection(closedConnection, thisConnectionID);
+          if (closedConnection != null) {
```
jd0-sag: Add a comment explaining how this can happen since it is not obvious.

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughServerProcess.java
```diff
@@ -240,7 +240,9 @@ public void shutdown() {
       this.notifyAll();
     }
     try {
-      this.serverThread.join();
+      if (this.serverThread != null) {
```
jd0-sag: Add a comment explaining how this can happen since it is not obvious.

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughEntityRef.java
```diff
@@ -119,7 +119,6 @@ public void create(C configuration) throws EntityNotProvidedException, EntityAlr
   public C reconfigure(C configuration) throws EntityException {
     // Make sure that we have a service provider.
     if (null != this.service) {
-      getWriteLock();
```
jd0-sag: It is probably worth adding a comment here, regarding why the write lock isn't required, if it is otherwise a common pattern.

PR #52, opened 2016-03-07
Proposal for getWithTimeout()

mathieucarbou:
We are now using getWithTimeout(). Sadly our api tests are failing now because the method is not implemented.

@jd0-sag @myronkscott : Could it be possible to redirect it to get() for the moment ?


jd0-sag:
The better answer would be to just implement getWithTimeout() since get() has different semantics.

It would be almost a copy-paste except that it requires a different wait() call and some accounting for time waited, as far as I can tell (or at least the knowledge that it could return for a reason other than a notify).

mathieucarbou:
@jd0-sag : ok! do you want me to try implementing it in this PR ? Or you can do it ? 
I fear there is some logic that I could miss so I think it could be better if you do it ?
Let me know what would be the quickest :-)

myronkscott:
I am working on several smaller tasks so I should be able to do this in the next day or two.

PR #60, opened 2016-03-25
Fix #14 - MessageCodec API changes

### LINKED ISSUE #14: EntityClientService requires getMessageCodec and client-side codec interface
MessageCodec currently provides translation between byte[] used on the wire and high-level types used within the server's entity implementation.

Introducing a client-side analogue of this interface would make it more obvious how the 2 sides are connected and would further split out the wire protocol details from the business logic of the client-side entity.

Note that this would require that the underlying EntityMessage and EntityResponse types move into a new "common entity API" module (we currently have only client or server, no common) since they would be the root interfaces used on either side of the wire.

Additionally, as it is likely that an implementation would want to use one implementation for both codec interfaces, we would need to ensure that the name of methods do not conflict, between the 2.

ramsai1729:
Introducing new entity-common-api module

	MessageCodec provides cleaner API to encode/decode EntityMessage
	and EntityResponse and its been moved to entity-common-api for
	client-side usage as well

	Segregated encode/decode API for Sync Messages into
	SyncMessageCodec and will be used on server-side

	Follow up changes for Passthrough and test-entity


entity-client-api/src/main/java/org/terracotta/entity/EntityClientService.java
```diff
@@ -61,4 +62,12 @@
    * @return entity
    */
   T create(EntityClientEndpoint endpoint);
```
mathieucarbou: Why not `EntityClientEndpoint<M, R>` ?

test-entity/client/src/main/java/org/terracotta/TestEntityClientServer.java
```diff
@@ -45,4 +46,9 @@ public TestEntityConfig deserializeConfiguration(byte[] configuration) {
   public TestEntity create(EntityClientEndpoint endpoint) {
```
mathieucarbou: `EntityClientEndpoint<TestMessage, TestResponse>` ?

entity-client-api/src/main/java/org/terracotta/entity/EntityClientService.java
ramsai1729: Fixed now, Thanks!

PR #61, opened 2016-03-30
Fix #55, Fix #56 - Add PlatformServer object.  

### LINKED ISSUE #55: Define "PlatformServer" information for monitoring
The IMonitoringProducer implementations expect more information to describe the server to which they are attached.

We need to find somewhere under `/platform` where we should put a node with the following kind of data:
`PlatformServer {`
`String hostName; // matches xml config`
`String bindAddress; // matches xml config`
`int bindPort; // matches xml config`
`int groupPort; // matches xml config`
`String state; // taken at runtime from servers`
`String version; // taken at runtime from servers`
`String buildId; // taken at runtime from servers`
`long startTime; // taken at runtime from servers`
`String hostAddress;`
`long activeTime;`
`}`

### LINKED ISSUE #56: Re-parent monitoring data under a server node to support multiple servers in a stripe
The monitoring data currently passed to IMonitoringProducer is per-server.  This creates a limitation in terms of how data from passive servers can be accessed.

While there are some ideas for how to collect this data, on the active, that implementation is still a ways off.

We know, however, that this will result in the tree being re-parented under the server, in order to support multiple servers.  This gives each server its own monitoring namespace and also implicitly connects any information in the tree to the server where it is applicable.

In order to make the data shape consistent with that eventual re-parenting, we should re-parent even the single-server information we provide, in order to show that level of indirection.

myronkscott:
Add more information to the PlatformServer and reparent to allow reporting of all the servers in the group.


monitoring-support/src/main/java/org/terracotta/monitoring/PlatformServer.java
```diff
@@ -0,0 +1,99 @@
...
+  private final int groupPort;
+  private final String version;
+  private final String build;
+  private final long startTime;
+
+  public PlatformServer(String serverName, String host, String bindAddress, int bindPort, int groupPort, String version, String build, long startTime) {
+    this.serverName = serverName;
+    this.hostName = host;
+    try {
+      this.hostAddress = InetAddress.getByName(host).getHostAddress();
```
mathieucarbou: @jd0-sag explained me that the voltron tree will contain all servers within a stripe. So I guess that if I am on the active, PlatformServer objects from passive are not created from this constructor ?

myronkscott: OK.  Will fix this to make it a pure data object

mathieucarbou:
@jd0-sag @myronkscott works for me :-)

monitoring-support/src/main/java/org/terracotta/monitoring/PlatformMonitoringConstants.java
```diff
@@ -33,6 +33,10 @@
    */
   public static final String PLATFORM_ROOT_NAME = "platform";
   /**
+   * The name of the node in the tree which is the parent to all nodes representing connected servers.
+   */
+  public static final String SERVER_ROOT_NAME = "servers";
+  /**
```
mathieucarbou: I think you would need that also to be consistent with the other constants:

`public static final String[] SERVERS_PATH = new String[]{PLATFORM_ROOT_NAME, SERVERS_ROOT_NAME};`

Issue #59, opened 2016-03-23
ServiceProvider should not extend Closeable


jd0-sag:
ServiceProviders live for the entire duration of the server so they can never be closed, nor do we have a definition of what that would mean.

This extension appears extraneous and should be removed.

jd0-sag:
While this is an API change, it is a very simple one, so I will look into the impact it will have and potentially add a commit to address it as part of the PR to address #58, later today.

Issue #58, opened 2016-03-23
MessageCodec.serializeForSync should operate on EntityMessage


jd0-sag:
The uses of MessageCodec for sync messages appears to be incorrect:  it expects to serialize a response and then deserialize a message when these are 2 sides of the same pipe, not a message-response round-trip.

Given that the deserialized message is passed to the entity invoke(), and EntityMessage is probably most appropriate.  The eventual plan is to use a specific type for messages related to sync.

Current:
`EntityResponse -> serialize -> byte[] -> (pipe) -> byte[] -> deserialize -> EntityMessage`
Proposed:
`EntityMessage -> serialize -> byte[] -> (pipe) -> byte[] -> deserialize -> EntityMessage`

jd0-sag:
I will be posting a PR including this change, today.  It is an API change, so downstream consumers need to sync up with us, but it is a trivial one.

PR #66, opened 2016-04-07
Fix passive sync codec generics and remove closeable on ServiceProvider

jd0-sag:
This addresses #58 and #59.


chrisdennis:
Makes sense to me... as of now the changes to accommodate this should be pretty easy to perform so now seems like a good time.

PR #68, opened 2016-04-20
client port is statically incremented for passthrough server

mathieucarbou:
to avoid having the same client port sequence when using 2 mocked clusters at the same time.

We have a test failing because we have a duplicate local port used for 2 separate cluster instances.

CC @jd0-sag @myronkscott @anthonydahanne


anthonydahanne:
LGTM

PR #71, opened 2016-04-26
Add Client PID to PlatformConnectedClient

ramsai1729:
This PR adds Client PID to PlatformConnectedClient and ClientDescriptor to PlatformClientFetchedEntity. 

With these changes server entities can look up PlatformConnectedClient like this 

1. Server entity looks up **PlatformClientFetchedEntity** using **ClientDescriptor** by iterating through list of **PlatformClientFetchedEntity** available in platform monitoring tree
2. Look up **PlatformConnectedClient** using **clientIdentifier** from **PlatformClientFetchedEntity** found in step 1

### DIFF
--- a/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughServerProcess.java
+++ b/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughServerProcess.java
@@ -466,7 +466,7 @@ public class PassthroughServerProcess implements MessageHandler {
               // Record that this entity has been fetched by this client.
               String clientIdentifier = clientIdentifierForService(sender.getClientOriginID());
               String entityIdentifier = entityIdentifierForService(entityClassName, entityName);
-              PlatformClientFetchedEntity record = new PlatformClientFetchedEntity(clientIdentifier, entityIdentifier);
+              PlatformClientFetchedEntity record = new PlatformClientFetchedEntity(clientIdentifier, entityIdentifier, clientDescriptor);
               String fetchIdentifier = fetchIdentifierForService(clientIdentifier, entityIdentifier);
               PassthroughServerProcess.this.serviceInterface.addNode(PlatformMonitoringConstants.FETCHED_PATH, fetchIdentifier, record);
             }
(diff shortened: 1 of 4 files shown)


mathieucarbou:
I have several questions about fe09823: 

1. What `ClientDescriptor.getClientIdentifier()` represents ? `<PID>@<ADDR>:<LOGICAL-CONNECTION-UUID>` ? Or just the logical connection name/uuid ? Or anything user-defined ?

2. Where are also the set of tags ?

3. How a developer can pass its own tags and client identifier when connecting ?

Thanks :-)

CC other M&M / Ehcache / Server people: @chrisdennis @ljacomet @anthonydahanne @jd0-sag

jd0-sag:
We probably want the ClientDescriptor to be stored as the actual ClientDescriptor type, since that is what it is meant to be compared against for connecting an entry in the tree with a client as it appears in connected(), disconnected(), or invoke().  The user can't act on this, only compare it against other instances since we state that hashcode() and equals() have reasonable implementations.

It also isn't yet exposed in the tree, by these changes.  We probably want to add it to PlatformClientFetchedEntity (although that is just my first thought).

The PID should probably be a long so that we can use the ProcessHandle, directly, once our baseline bumps up to Java9 (minimally, this makes us consistent with any expectations created by the introduction of that class).

@mathieucarbou This change just adds those platform intrinsic pieces of information which we had settled on, yesterday.  Beyond that, the issue of tags has been deferred (since there isn't agreement on what they would look like or where in the stack they would be exposed) and the ability to pass a "logical name" from the client will be added in a later change since, while it also needs to appear in PlatformConnectedClient, it is client-originating data which changes an API and was still in flux, earlier today.

mathieucarbou:
The difference to keep in mind for M&M between `PlatformClientFetchedEntity` and `PlatformConnectedClient` is that if some information are moved to the `PlatformClientFetchedEntity`, it means we cannot access them until an entity is fetched.

That means that in M&M we won't be able to display any client information about connections made for entity creation / destroy (i.e. __ehcache-cli__) or for example a client that connects, then crashes before fetching an entity.

Basically we won't be able to show any clients in M&M until a fetch happens.

jd0-sag:
The reason why `ClientDescriptor` would be added to `PlatformClientFetchedEntity` is that it is only defined within the context of a fetch.  It is akin to a file descriptor: it is only defined within the context of an open file living within a specific process.

The other information, such as PID and logical connection name, can live in `PlatformConnectedClient` since that information is associated with the connection, no matter what it is doing.  This PID is a platform-provided piece of data while the logical connection name is something we will add to the client-side connection API in a future change.

PR #76, opened 2016-04-29
Fix #75

### LINKED ISSUE #75: PassthroughServerProcess does not expose client's PID
`PlatformConnectedClient` returns -1 with the passthrough system.

mathieucarbou:


jd0-sag:
Do you want each connection to look like it is coming from the same process or did we want to fake this up along the same lines as what we are doing for client-side remote ports?

mathieucarbou:
From the same process is ok - at least for now. It's quite hard to assert in tests when values are randomly generated. For the PID, I don't consider it random since we can know in advance what it will be. So we are able to get also the PID to verify that the topology we have is the expected one.

The only thing that should be generated is the client port number, as it is now, and it is a value that we cannot event know in advance.

jd0-sag:
As long as you guys are ok with it being the same for all connections, that works for me.

PR #84, opened 2016-05-10
Proposal for #83 for test lib

### LINKED ISSUE #83: Fix tree exposures vs server entity callbacks
We would need to setup a tree structure that has some references to the current connection. Thus we would need access to the `PlatformConnectedClientObject`, and `PlatformFetchedEntity` object from the server entity callbacks (by using the `ClientDescriptor`)

```
connected(ClientDescriptor clientDescriptor)
disconnected(ClientDescriptor clientDescriptor)
```

So the idea would be that platform exposes in the tree these 2 objects __before__ `connected()` is called and remove then __after__ `disconnected()` is called.

mathieucarbou:
This fixes #83 for the test lib, but still tc-core needs to be changed.


jd0-sag:
We will need to make some changes to doc and then coordinate the change with core and some internal tests so it is probably best to leave this for us.

anthonydahanne:
cloudbees, test this please

mathieucarbou:
@jd0-sag : do you have an ETA for this ? We have 2 PR waiting for it, to finish a user story task. 
Note: I've tested the passthrough change, it works fine with our tests.

myronkscott:
I am currently working on this.  Will be merged in a separate PR

PR #87, opened 2016-05-17
Add PlatformService API

ramsai1729:


jd0-sag:
You should add a minimal passthrough implementation, both for testing portability, but also to better see the data which is provided.

PR #91, opened 2016-05-20
Fix leaked PassthroughServer pseudo-connection

jd0-sag:
* this is now life-cycled along with the server
* also improved use of connection thread names


jd0-sag:
@cljohnso This is ready to merge unless you have any concerns with it (no interaction with assumptions in other repos).

cljohnso:
I can't imagine anyone depending on the internal connection behavior ...

PR #93, opened 2016-05-24
Fix #80: Created "passthrough" URI and ConnectionService

### LINKED ISSUE #80: Create ConnectionService implementation for Passthrough
Uses of the passthrough test environment are currently limited to the cases where the Connection object can be accessed, directly.  Many applications would rather operate on a URI since that is how a more general-purpose configuration is performed, etc.

Passthrough could be used to test these applications if we provided a URI scheme and corresponding ConnectionService.

jd0-sag:
-for a PassthroughServer to be reachable via the PassthroughConnectionService, it needs to be registered in the global PassthroughServerRegistry
-the PassthroughServerRegistry is a global because there is no way to pass additional data into the ConnectionFactory to help resolve the server instance (it allows a way to hook into a synthetic name registry - emulating the network)

### DIFF
--- a/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughClusterControl.java
+++ b/entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughClusterControl.java
@@ -18,13 +18,18 @@
  */
 package org.terracotta.passthrough;
 
+import java.net.URI;
+
 import org.terracotta.connection.Connection;
+import org.terracotta.connection.ConnectionException;
+import org.terracotta.connection.ConnectionFactory;
 
 
 /**
  * The implementation used to control the passthrough testing cluster.
  */
 public class PassthroughClusterControl implements IClusterControl {
+  private final String stripeName;
   // We track the "original" server state for tear-down.
   private final PassthroughServer originalActiveServer;
   private final PassthroughServer originalPassiveServer;
@@ -32,13 +37,16 @@ public class PassthroughClusterControl implements IClusterControl {
   // The active we are currently using can change on restart.
   private PassthroughServer activeServer;
 
-  public PassthroughClusterControl(PassthroughServer activeServer, PassthroughServer passiveServer) {
+  public PassthroughClusterControl(String stripeName, PassthroughServer activeServer, PassthroughServer passiveServer) {
+    this.stripeName = stripeName;
     // The active cannot be null but the passive can be.
     Assert.assertTrue(null != activeServer);
     this.originalActiveServer = activeServer;
     this.originalPassiveServer = passiveServer;
     // We set the changing active server to be the original.
     this.activeServer = activeServer;
+    PassthroughServer previous = PassthroughServerRegistry.getSharedInstance().registerServer(this.stripeName, activeServer);
+    Assert.assertTrue(null == previous);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


jd0-sag:
@mathieucarbou Can you take a look at this and let me know if it is what you had in mind?

mathieucarbou:
I will :-) thanks!

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughClusterControl.java
```diff
@@ -18,27 +18,35 @@
...
+  private final String stripeName;
   // We track the "original" server state for tear-down.
   private final PassthroughServer originalActiveServer;
   private final PassthroughServer originalPassiveServer;
 
   // The active we are currently using can change on restart.
   private PassthroughServer activeServer;
 
-  public PassthroughClusterControl(PassthroughServer activeServer, PassthroughServer passiveServer) {
+  public PassthroughClusterControl(String stripeName, PassthroughServer activeServer, PassthroughServer passiveServer) {
```
mathieucarbou: Since there's an api change, would you like to also change the class name to sort of PassthroughStripeControl ? Since a cluster is composed of several stripes and this class just handles a stripe.

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughConnectionService.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  public boolean handlesURI(URI uri) {
+    return SCHEME.equals(uri.getScheme());
+  }
+
+  @Override
+  public Connection connect(URI uri, Properties properties) throws ConnectionException {
+    Connection connection = null;
+    String serverName = uri.getHost();
```
mathieucarbou: Isn't it the stripe name instead ?
```
   public Connection createConnectionToActive() {
       URI uri = URI.create("passthrough://" + this.stripeName);
```

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughClusterControl.java
jd0-sag: It is more just so it syncs up with the interface name (which _can_ extend beyond one stripe).

Do you use the PassthroughClusterControl, in your tests?

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughConnectionService.java
jd0-sag: That is basically an implementation detail, and I am not sure that I really like it (it fits within how the PassthroughServerProcess currently works but I actually think that is somewhat confusing - ideally, this would work more like the real network and not make as many assumptions about the shape of the stripe - this also may end up changing to support multiple passives).

mathieucarbou:
@jd0-sag : this is approximately what we did.

In our case we created a `ConnectionService` linked to a `ClusterControl` thanks to a class we called: `PassthroughConnectionManager`.

In our API test, we can then mock dynamically any URI schemes to forward them to the passthrough like this:

```
PassthroughConnectionManager.getInstance().mock("terracotta")
// run the test
PassthroughConnectionManager.getInstance().unmock()
```

We have also these methods:

```
IClusterControl ctl = passthroughConnectionManager.getOrCreateClusterControl();
ServiceRegistry voltronRegistry = passthroughConnectionManager.getServiceRegistry();
```

We also had to hack the classloader to be sure the Passthrough Connection Service comes first in the classpath... So it's quite ugly.

If this gets merged I'll remove what we did because after looking at our tests, we could just change our connection scheme to passthrough:// I think it should work.

mathieucarbou:
@jd0-sag : BTW, do you want me to merge or it needs to be in sync with something else ?

entity-test-lib/src/main/java/org/terracotta/passthrough/PassthroughClusterControl.java
mathieucarbou: Actually we are using it to create another Connection to the same stripe to simulate another client. But we could have just used the ConnectionService with the right scheme. So the answer is: yes, but I'll change that :-)

PR #97, opened 2016-05-27
Fix #95: Call destroy() on the entity, after removing it

### LINKED ISSUE #95: PassthroughServerProcess.destroy does not call CommonServerEntity.destroy
When `destroy` or `tryDestroy` is called on a client-side entity, one expects the `destroy` method on the server-side entity to be called.  This is not happening when using `PassthroughServer` -- the call to `CommonServerEntity.destroy` is missing from `PassthroughServerProcess.destroy`.

jd0-sag:
This shouldn't be merged yet - there is another PR we want to see merged, first.


jd0-sag:
This is now ready to merge - it has been rebased.

PR #99, opened 2016-05-31
added toString() to help debugging when dumping the monitoring tree

mathieucarbou:


mathieucarbou:
@anthonydahanne : FYI, also added this one so that we know what's going on in the voltron tree in debug mode

PR #104, opened 2016-06-02
Re-word the ConcurrencyStrategy doc

jd0-sag:
-this is primarily to make the rule that the implementation must return the same key, for a given message, no matter how many times it is called, more obvious
-also updated some of the other wording to reflect our current types, etc


jd0-sag:
This is just a small concern but let me know what you guys think of this wording.

PR #111, opened 2016-06-17
Add multi passive support in passthrough framework

ramsai1729:
This PR adds support for multiple passives in passthrough framework

- PasshtroughClusterControl controls server start/stop and active transition 
- Removes strong coupling between active and passive 
- Adds crashActive method to IClusterControl to test passive transition

### DIFF
--- a/entity-test-lib/src/main/java/org/terracotta/passthrough/IClusterControl.java
+++ b/entity-test-lib/src/main/java/org/terracotta/passthrough/IClusterControl.java
@@ -42,6 +42,20 @@ public interface IClusterControl {
   public void waitForActive() throws Exception;
 
   /**
+   * Terminates current active of this Stripe
+   *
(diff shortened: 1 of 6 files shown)


jd0-sag:
I think that the crux of this change is very good (it does a lot to clean up how active-passive relationships work, in passthrough).

I want you to walk me through it, though, to ensure that I understand the general ideas behind the design and the difference between stopActive and restartActive.  I need to add some new capabilities to IClusterControl, including things like stopping the active, so I can probably just build on top of the direction you were taking this, so long as I understand how you are looking at it.

On a less important note, there are a few gratuitous whitespace changes in otherwise untouched areas of the code, which you probably want to strip from the change, just to avoid version history perturbations.

Let me know if you are free to talk about this, and any implications it has for galvan, on Monday.

PR #125, opened 2016-07-08
Removed entity-test-lib as it now moved to tc-passthrough-testing

jd0-sag:
-also removed JUnit references from poms

This module is being moved to https://github.com/Terracotta-OSS/tc-passthrough-testing


jd0-sag:
(self-merging since this was discussed with @myronkscott earlier, today)

PR #126, opened 2016-07-13
EntityRef API contract updates

ljacomet:
~~The goal of this PR is not to be merged but to serve as a support for updates to the `EntityRef` API contract.~~ Changes _accepted_

The main point is the way `create`, `fetchEntity` and `destroy` interact between each others or when invoked concurrently.

### DIFF
--- a/connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
+++ b/connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
@@ -26,27 +26,48 @@ import org.terracotta.exception.EntityVersionMismatchException;
 
 
 /**
- * The EntityRef is a reference to where an entity could potentially exist.  That is, it is created to refer to a specific
- * entity type and name, although there may not yet be an entity with that type and name.
- * 
- * The EntityRef is used to manage the server-side entity's life cycle in that it is how the entity is either created or
- * destroyed.
- * 
- * It is also used as the path through which an actual instance can be fetched for use on the client.
- * 
+ * The EntityRef is a reference to where an entity could potentially exist.
+ * <P>
+ *   That is, it is created to refer to a specific entity type and name, although there may not yet be an entity with
+ *   that type and name.
+ * </P>
+ * <P>
+ *   The EntityRef is used to manage the server-side entity's life cycle in that it is how the entity is either created
+ *   or destroyed.
+ * </P>
+ * <P>
+ *   It is also used as the path through which an actual instance can be fetched for use on the client.
+ * </P>
+ *
  * @param <T> The entity type underlying this reference.
  * @param <C> The configuration type to use when creating this entity.
  */
 public interface EntityRef<T extends Entity, C> {
   /**
-   * Create the entity with the given configuration.
+   * Creates the entity with the given configuration.
+   * <P>
+   *   If the entity already exists, the {@link EntityAlreadyExistsException} is  thrown immediately, whether or not
+   *   the entity is fetched on any client.
+   * </P>
+   * <P>
+   *   If a {@code create} is invoked while another {@code create} is happening, it will block until either:
+   *   <UL>
+   *     <LI>the other create completes and then throw an {@code EntityAlreadyExistsException}</LI>
+   *     <LI>the other create fails and then it will (try to) perform the create as well</LI>
+   *   </UL>
+   *   The contract provides guarantee that when multiple clients race to {@code create}, and in the absence of any
+   *   {@link #destroy()} or server failure, when the {@code create} method completes (returns or throws), a subsequent
+   *   {@link #fetchEntity()} will not throw an {@link EntityNotFoundException}.
+   * </P>
    *
    * @param configuration configuration to be applied to the entity
-   * @throws EntityNotProvidedException The service providing T doesn't exist on either the client or the server
+   *
+   * @throws EntityNotProvidedException The service providing {@code <T>} doesn't exist on the server
    * @throws EntityAlreadyExistsException An entity with this type and name already exists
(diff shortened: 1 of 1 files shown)


connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
```diff
@@ -26,27 +26,48 @@
...
+   *     <LI>the other create completes and then throw an {@code EntityAlreadyExistsException}</LI>
+   *     <LI>the other create fails and then it will (try to) perform the create as well</LI>
+   *   </UL>
+   *   The contract provides guarantee that when multiple clients race to {@code create}, and in the absence of any
+   *   {@link #destroy()} or server failure, when the {@code create} method completes (returns or throws), a subsequent
+   *   {@link #fetchEntity()} will not throw an {@link EntityNotFoundException}.
+   * </P>
    *
    * @param configuration configuration to be applied to the entity
-   * @throws EntityNotProvidedException The service providing T doesn't exist on either the client or the server
```
ljacomet: Removed references to the client because without a client service there should be no way to get to the `EntityRef` in the first place.

connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
```diff
@@ -56,34 +77,33 @@
...
-   * @throws EntityNotFoundException No entity with this type and name could be found
-   */
-  void destroy() throws EntityNotProvidedException, EntityNotFoundException;
+   * Destroys the entity pointed to by this reference if it is not fetched on any client.
+   * <P>
+   *   If the entity is fetched on any client, this method returns false.
+   * </P>
+   *
+   * @return {@code true} if the entity was destroyed, {@code false} if there are fetched entities on any client
+   * preventing destruction
```
mathieucarbou: Then why not just throwing a sort of `EntityBusyException` that you could re-use in Ehcache ?

If a boolean return is required, I would prefer to have `boolean tryDestroy()` then :-)

ljacomet: Given that we are building mostly on top of non blocking ops, I think the defaults should have the same characteristics. Now we could use an exception ... but that seems weird in some ways.

connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
jd0-sag: It can still throw if there is no provider on the server, though.  This is effectively an incomplete server installation.

jd0-sag:
Were you going to add the blocking versions of any of these calls, at this point, or were we going to first figure out the cases where those actually make sense and see if we even want them?

connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
myronkscott: From a symmetry (to create) standpoint EntityBusyException or EntityReferencedException seems interesting.  But from the standpoint of what will happen the majority of the time this is called (failure) maybe returning a boolean is proper so the caller can enter some external wait condition.

ljacomet:
@jd0-sag I was thinking about addressing what we identified as needed through Ehcache development. As @myronkscott mentioned, the client can use _non-blocking_ ops to perform a _blocking_ one.

connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
```diff
@@ -26,27 +26,48 @@
...
+ *
  * @param <T> The entity type underlying this reference.
  * @param <C> The configuration type to use when creating this entity.
  */
 public interface EntityRef<T extends Entity, C> {
   /**
-   * Create the entity with the given configuration.
+   * Creates the entity with the given configuration.
+   * <P>
+   *   If the entity already exists, the {@link EntityAlreadyExistsException} is  thrown immediately, whether or not
```
ljacomet: Condition that indicates we do not want the operation to block when the entity is already fetched on any client.

connection-api/src/main/java/org/terracotta/connection/entity/EntityRef.java
```diff
@@ -26,27 +26,48 @@
...
+   * </P>
+   * <P>
+   *   If a {@code create} is invoked while another {@code create} is happening, it will block until either:
+   *   <UL>
+   *     <LI>the other create completes and then throw an {@code EntityAlreadyExistsException}</LI>
+   *     <LI>the other create fails and then it will (try to) perform the create as well</LI>
+   *   </UL>
+   *   The contract provides guarantee that when multiple clients race to {@code create}, and in the absence of any
+   *   {@link #destroy()} or server failure, when the {@code create} method completes (returns or throws), a subsequent
+   *   {@link #fetchEntity()} will not throw an {@link EntityNotFoundException}.
```
ljacomet: This paragraph explains the reasoning behind the list just above it - a sensible concurrent contract. Here I am interested in pitfalls that this _block while other finishes create_ may cause.

ljacomet:
I believe there is agreement on this proposal and thus it could be merged.
The only question is whether the creation of an issue that could be referenced from downstream repositories when implementing / adapting to the changes proposed here bring anything vs. a reference to the PR itself.

jd0-sag:
My thinking is that they could reference the PR, directly, since it already has the context of the discussion.  The commit and PR should probably be named in terms of their conclusive changes, though.

ljacomet:
@jd0-sag Given your comment, realised that having an issue is better as it allows the commit to reference it as having it reference its PR would be weird.
Also updated the commit comment and that PR title to match.

jd0-sag:
Closed by #141 (a rebased version of this same commit)

PR #128, opened 2016-07-14
Fix #127 - Add support to introspect EntityExceptions data

### LINKED ISSUE #127: Add support to introspect EntityExceptions data
- Add support to introspect EntityExceptions data
- Add a constructor to set cause for EntityAlreadyExistsException, EntityNotFoundException, EntityNotProvidedException, EntityVersionMismatchException and PermanentEntityException 

This support is needed for Terracotta-OSS/terracotta-core#253 where we need re-create the corresponding exception on client side

ramsai1729:
This PR adds
- support to introspect EntityExceptions data
- a constructor to set cause for EntityAlreadyExistsException, EntityNotFoundException, EntityNotProvidedException, EntityVersionMismatchException and PermanentEntityException


jd0-sag:
Since this is a purely additive change (to objects, not interfaces), I will merge this change first.

The other PR can be merged later today, once we have a better idea of how to coordinate the next release (while these don't NEED to be lock-step, it would be nice to get a concrete new release).

PR #129, opened 2016-07-14
Fix #57 - Rename ServerEntityService class to EntityServerService

### LINKED ISSUE #57: EntityClientService and ServerEntityService should be named consistently
These classes are the analogues of each other, existing on the client and server, respectively.

Their names are slightly different, as they evolved to these points differently, which makes them hard to search for.

We should change `ServerEntityService` to `EntityServerService` so that they both have the same shape and one is more obvious to find, starting from the other.

ramsai1729:


jd0-sag:
Do you have the corresponding core changed, required by this?  We need to do some coordination to get this release out but you don't seem to be online.

jd0-sag:
This change will need a rebase and we can coordinate its merge with the release, later today.

PR #133, opened 2016-07-28
(READY TO MERGE) Fix #131: Expanded ServiceProvider Javadoc

### LINKED ISSUE #131: Javadoc the thread safety guarantees expected from ServiceProvider implementers
The thread safety guarantees expected from the implementer of `ServiceProvider` interface need to be javadoc'ed. Especially for the _getService()_ method.

jd0-sag:
-this includes an expanded description of threading concerns, when/why certain methods are called, and what a "consumerID" is

NOTE:  This should be harmless to merge but we want to coordinate rebases with outstanding pull requests before we know the order.

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
@@ -22,46 +22,79 @@ import java.util.Collection;
 
 
 /**
- * Service Provider which are provided at the platform level.  A service provider is responsible for instantiating the
- * services to interact with its underlying implementation in response to platform requests.  While the services are
- * generally used by entities, they are also used by the platform, itself.
+ * A service provider is responsible for instantiating the services to interact with its underlying implementation in response
+ *  to platform requests.  While the services are generally used by entities, they are also used by the platform, itself.
+ * 
+ * It is possible for a given ServiceProvider implementation to expose multiple different service types, as returned by
+ *  {@link #getProvidedServiceTypes()}.  When an entity requests a service instance, the ServiceConfiguration it provides is
+ *  used to determine which ServiceProvider instances to check.
+ * NOTE:  Currently, it is a failure for more than one ServiceProvider to produce a service instance for the same
+ *  ServiceConfiguration.  It is considered a fatal server misconfiguration due to the ambiguity.
+ * 
+ * Note that there are no threading restrictions on how the methods in this interface will be called.  This means that any
+ *  implementation which has shared state must ensure that it can manage this state in a thread safe way (subject to whatever
+ *  thread safety its own state requires).
  */
 public interface ServiceProvider {
+  /**
+   * PLATFORM_CONSUMER_ID is the consumerID used by the platform, itself, when it wants to interact with services.  Normal
+   *  entity instances use positive consumerID values.
+   */
+  public final long PLATFORM_CONSUMER_ID = 0; 
 
   /**
-   * The platform configuration based on which the Service provider can choose to initialize itself.
-   *
-   * @param configuration platform configuration
-   * @return true if provider has successfully been initialized with the provided configuration
+   * The platform configuration based on which the receiver can choose to initialize itself.
+   * This is called once on each provider instance, at server start-up time, before any other methods have been called on the
+   *  instance.
+   * Note that it is possible for there to be multiple instances of the same ServiceProvider if the server had multiple
+   *  configuration stanzas handled by the same ServiceProvider class.  As the entities requesting services do not have direct
+   *  access to the set of initialized ServiceProviders, they do not require special accounting for this possibility.  The
+   *  entities can still only select a ServiceProvider via a matching ServiceConfiguration passed to
+   *  {@link #getService(long, ServiceConfiguration<T>)}.
+   * 
+   * @param configuration The configuration for this ServiceProvider instance.
+   * @return True if provider has successfully been initialized with the provided configuration, false if this instance
+   *  shouldn't be registered for the given configuration (it will be discarded).
    */
   boolean initialize(ServiceProviderConfiguration configuration);
 
   /**
-   * Get an instance of service from the provider.
-   *
-   * @param consumerID The unique ID used to name-space the returned service
-   * @param configuration Service configuration which is to be used
-   * @return service instance
+   * Get an instance of a service from the provider.
+   * Note that this method can be called any number of times, over the life of a ServiceProvider instance, potentially
+   *  concurrently or multiple times for the same consumerID.  If the implementation wishes to intern, or otherwise track,
+   *  the instances it returns, it will need to ensure its approach is robust against concurrent calls.
+   * Note that the consumerID is a unique number for each entity instance which is used to effectively name-space the service
+   *  instances under the entities.  An entity instance will have the same consumerID for its entire life (including after
+   *  restart or fail-over).  The value is always non-negative.
+   * The receiver is expected to return null if it is not able to create a service with the given ServiceConfiguration.
+   * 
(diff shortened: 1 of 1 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,62 @@
...
-
   /**
-   * The platform configuration based on which the Service provider can choose to initialize itself.
-   *
-   * @param configuration platform configuration
-   * @return true if provider has successfully been initialized with the provided configuration
+   * The platform configuration based on which the receiver can choose to initialize itself.
+   * This is called once on each provider instance, at server start-up time, before any other methods have been called on the
+   *  instance.
+   * Note that it is possible for there to be multiple instances of the same ServiceProvider if the server had multiple
```
ljacomet: What happens if that is the case? How do users of `ServiceRegistry` choose the one they need?

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,62 @@
...
-   * Get an instance of service from the provider.
-   *
-   * @param consumerID The unique ID used to name-space the returned service
-   * @param configuration Service configuration which is to be used
-   * @return service instance
+   * Get an instance of a service from the provider.
+   * Note that this method can be called any number of times, over the life of a ServiceProvider instance, potentially
+   *  concurrently or multiple times for the same consumerID.  If the implementation wishes to intern, or otherwise track,
+   *  the instances it returns, it will need to ensure its approach is robust against concurrent calls.
+   * Note that the consumerID is a unique number for each entity instance (consumerID 0 represents the platform, itself) which
```
ljacomet: Could we have a constant reference there instead of a magic number?
Is there any rule on the sign of that consumerID?

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,62 @@
...
-   *
-   * Generally platform calls this method during platform initialization so there won't be any entities using
-   * underlying services but ServiceProviders need to make sure that no entities (with the exception of platform whose consumerID is 0) using underlying services
-   *
-   * If there are any failures when clearing state, this method should inform Platform by throwing {@link ServiceProviderCleanupException}
+   * Clears up state for this ServiceProvider including any persisted state.
+   * This is called, by the platform, when it wishes to explicitly clear the server's state (both in-memory and persistent).
+   * Generally, the platform calls this method during platform initialization so there won't be any entities using the
+   *  underlying service instances.  In a more concrete sense, no services instances created before this call are considered
+   *  valid after the call, except for those created for consumerID 0 (which is the platform-internal consumer).
```
ljacomet: Wait, does that mean that clear should not clear state linked to `consumerID 0`?

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
myronkscott: Based on the ServiceConfiguration that is passed in by the entity when asking for a service.  Each ServiceProvider will be asked if it can satisfy the ServiceConfiguration.  For example, a Service implementor can use a naming scheme and pass a unique name with the ServiceConfiguration.

jd0-sag: I will add a note that the multiple instances aren't directly visible to the calling entity (since they don't directly access the providers).

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
jd0-sag: Good call.  I wasn't sure if this is the right place for it but it probably is (and we can move it later, if need be).

I will also add a note that all consumerID values are non-negative.

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
jd0-sag: This confuses me, as well.  It probably makes more sense to remove this special case, say that all state should be cleared, and then ensure that that the implementation will re-request the service instance if it ever needs it (since calling a bizarre implementation detail a bug is preferable to making it a special-case in the API).

jd0-sag:
I updated in response to these suggestions.  Let me know what else should be considered.  Otherwise, we can figure out how to coordinate this with the other PR.

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,69 @@
...
+ *  to platform requests.  While the services are generally used by entities, they are also used by the platform, itself.
+ * 
+ * Note that there are no threading restrictions on how the methods in this interface will be called.  This means that any
+ *  implementation which has shared state must ensure that it can manage this state in a thread safe way (subject to whatever
+ *  thread safety its own state requires).
  */
 public interface ServiceProvider {
+  /**
+   * PLATFORM_CONSUMER_ID is the consumerID used by the platform, itself, when it wants to interact with services.  Normal
+   *  entity instances use positive consumeID values.
```
ljacomet: typo: consumerID
Will also look better if wrapped inside `{@code }` javadoc instruction

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,69 @@
...
-   * The platform configuration based on which the Service provider can choose to initialize itself.
-   *
-   * @param configuration platform configuration
-   * @return true if provider has successfully been initialized with the provided configuration
+   * The platform configuration based on which the receiver can choose to initialize itself.
+   * This is called once on each provider instance, at server start-up time, before any other methods have been called on the
+   *  instance.
+   * Note that it is possible for there to be multiple instances of the same ServiceProvider if the server had multiple
+   *  configuration stanzas handled by the same ServiceProvider class.  As the entities requesting services do not have direct
+   *  access to the set of initialized ServiceProviders, they do not require special accounting for this possibility.
```
ljacomet: I also find this comment on the `initialize` method to be a bit weird. Since that is lifecycle handled outside of this method and others here, this should move to class level javadoc. And there you could also explain how different `ServiceProvider` will be invoked one after another in case a specific one cannot provide its service because the `ServiceConfiguration` does not match.

entity-server-api/src/main/java/org/terracotta/entity/ServiceProvider.java
```diff
@@ -22,46 +22,69 @@
...
+   *  concurrently or multiple times for the same consumerID.  If the implementation wishes to intern, or otherwise track,
+   *  the instances it returns, it will need to ensure its approach is robust against concurrent calls.
+   * Note that the consumerID is a unique number for each entity instance which is used to effectively name-space the service
+   *  instances under the entities.  An entity instance will have the same consumerID for its entire life (including after
+   *  restart or fail-over).  The value is always non-negative.
+   * 
+   * @param consumerID The unique, non-negative, ID used to name-space the returned service for a specific consuming entity
+   *  (or the platform).
+   * @param configuration Service configuration which is to be used.
+   * @return The service instance (which may be either new or a cached/interned instance for this consumerID).
```
ljacomet: What do you return if a specific instance cannot provide the service and another instance should be providing it instead?

ljacomet:
Added another set of comments around lifecycle and multiple `ServiceProvider` instances of the same type.

As for the link with the other PR, I would suggest merging both PR and making sure the next terracotta-core version embarks both set of changes. If that is deemed as too big at once, then indeed some priority decision will need to happen.

Issue #135, opened 2016-08-10
Serializability of the data passed to IMonitoringProducer must be defined


jd0-sag:
Currently, `IMonitoringProducer` accepts data in the form of `Object` instances.  This works fine for objects which never leave the address space of the JVM.  However, this limits how the data can be used.  Specifically, it prevents the platform from being able to investigate ways of passing this data between servers on the same stripe, which will likely be required for effective monitoring of passive server state and entities.

Here are 3 suggestions for how to address this, to get the conversation started:

1. Change the value type from `Object` to `Serializable`.  This is simple but does fix the serialization approach which may appear sub-optimal to some use-cases.
1. Change the value type from `Object` to `byte[]`.  This maximizes the freedom of the serialization approach but pushing this burden into the services and entities does make their implementation quite cumbersome.
1. Require that a codec is passed into the monitoring service when it is looked up (as part of its `ServiceConfiguration`).  This works well on the encoding side but not so well on the decoding side where it isn't clear which codec would be used (although it may be possible to assume they are the same, between servers, for a given consumer ID).

jd0-sag:
@mathieucarbou @anthonydahanne @myronkscott @ramsai1729: Let me know what you think.

mathieucarbou:
I would be in favor of 1. As we discussed, this seems to be the easier one and then the platform would be responsible for the serialization, the way it wants to do it.

anthonydahanne:
I also like 1 for its simplicity.
As a voltron user, 2 does not sound appealing :-)
if 3 was chosen I'm pretty sure we would use a VoltronProxy-like layer to handle and keep in sync the codecs across nodes

jd0-sag:
For the most part, this change to Serializable is relatively straight-forward.  The exception to this appears to be in `PlatformClientFetchedEntity` since there is a `ClientDescriptor` as part of it.

My current thinking is to make this a `transient` instance variable for 2 reasons:

1. This is somewhat similar to the idea of a file descriptor in that deserializing it doesn't mean that you can use it in any meaningful or safe way
1. The only reason why we are likely to serialize this object (and this **MUST** be documented in the class) is if we are passing information from the passive back to the active and `PlatformClientFetchedEntity` instances aren't created on the passive (meaning that this is well-defined way of limiting the scope of what is probably a moot point)

PR #136, opened 2016-08-16
Fix #134: Added IMonitoringProducer.pushBestEffortsData

### LINKED ISSUE #134: Add a best-efforts "event push" method to IMonitoringProducer
The methods in `IMonitoringProducer` are effective when it comes to communicating opaque data in a "high-fidelity" way:  an implementation of the interface can keep the data set to a well-defined size even as events mutate it, due to the tree structure.

We need a way to communicate "low-fidelity" data, though.  This includes things like statistics or frequent state notifications which we would like to communicate but can be safely dropped without any impact on correctness.

The suggested interface is:
`
public void pushBestEffortsData(String[] path, Object data);
`

The underlying implementation is free to drop these messages in the cases where they are arriving more quickly than they can be consumed by some underlying process.  The implementation decides both how much data it will try to hold, before dropping, as well as the dropping strategy ("oldest", "most similar", or "most recent", for example).

jd0-sag:
-this provides a mechanism for a proder of monitoring data to rapidly send this data through a lower-overhead, less reliable channel
-the definition of the reliability of this mechanism is left up to the implementation

NOTE:  The implementation of #135 will be made in a different PR since it has an impact on some internal types which will require a little more testing.  It will likely end up being posted before the next release is cut, though.

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/monitoring/IMonitoringProducer.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/monitoring/IMonitoringProducer.java
@@ -28,4 +28,19 @@ public interface IMonitoringProducer {
    * @return True if the node was removed.  False if it or a parent couldn't be found.
    */
   public boolean removeNode(String[] parents, String name);
+
+  /**
+   * Makes a best-efforts attempt to push named data to the interface.  This method differs from the add/remove node methods in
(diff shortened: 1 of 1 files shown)


mathieucarbou:
Good to me!
@jd0-sag , @myronkscott : can I merge ?

PR #137, opened 2016-08-17
Fix #135: IMonitoringProducer values must be Serializable

### LINKED ISSUE #135: Serializability of the data passed to IMonitoringProducer must be defined

jd0-sag:
-this will ensure that future concerns of how to communicate monitoring data between active and passive servers will be possible
-this also required that the standard types defined in monitoring-support be made serializable
-of specific note, this means that PlatformClientFetchedEntity's ClientDescriptor had to be marked transient, since it can't be meaningfully deserialized (it is akin to deserializing a file descriptor).  These should appear in cases where this would matter, however, as these are not created by the passive
-for testing reasons, this also required the addition of hashCode() and equals() methods to these types (with the assumption that only this ClientDescriptor field can be null and only due to its transient nature)

### DIFF
--- a/monitoring-support/src/main/java/org/terracotta/monitoring/PlatformClientFetchedEntity.java
+++ b/monitoring-support/src/main/java/org/terracotta/monitoring/PlatformClientFetchedEntity.java
@@ -18,18 +18,29 @@
  */
 package org.terracotta.monitoring;
 
+import java.io.Serializable;
+
 import com.tc.classloader.CommonComponent;
 import org.terracotta.entity.ClientDescriptor;
 
 
 /**
  * A type which describes the connection between a client and an entity, created by a fetch.
+ * 
+ * WARNING:  The clientDescriptor is marked as transient since it can't be meaningfully deserialized and the cases where this
+ *  object will be used are also cases where it won't be serialized.  Be careful of using this object in these other cases.
  */
 @CommonComponent
-public class PlatformClientFetchedEntity {
-  public final String clientIdentifier;
-  public final String entityIdentifier;
-  public final ClientDescriptor clientDescriptor;
+public class PlatformClientFetchedEntity implements Serializable {
+  private static final long serialVersionUID = 4741752382657834201L;
+
+  public String clientIdentifier;
+  public String entityIdentifier;
+  public transient ClientDescriptor clientDescriptor;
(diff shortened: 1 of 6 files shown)


mathieucarbou:
If `ClientDescriptor` is transient, does it mean that if a failover occurs, we could have in the new active's monitoring tree some `PlatformClientFetchedEntity` objects having null `ClientDescriptor`s ?

More generally, since `PlatformClientFetchedEntity` represent a client that has fetched an entity, if a failover occurs, then, my guess is that all `PlatformClientFetchedEntity` should be removed and since the client will reconnect to the new active, new `PlatformClientFetchedEntity` will be created right ?

So... at the end... Even if `ClientDescriptor` is transient, and thus, can be null in `PlatformClientFetchedEntity`, does it make sense to replicate this object, and __more important question for us__: could we see a `PlatformClientFetchedEntity` object with a null `ClientDescriptor` (which is an invalid use case and will crash of course in our code) ?

Thanks!

jd0-sag:
I don't expect a null `ClientDescriptor` to be seen in any case, actually, since the passive doesn't create `PlatformClientFetchedEntity` instances and therefore the active won't see deserialized copies of them.

In the case of fail-over, the passive will become active and will repopulate the tree as the clients reconnect and reestablish their respect entity fetch states from their reconnect handshakes.

PR #146, opened 2016-09-06
(DISCUSS) Fix #144:  Added clarification regarding concurrency keys running life-cycle calls

### LINKED ISSUE #144: Document concurrency levels for Active and Passiver ServerEntity APIs
There have been recent questions as to what exactly are the concurrency guarantees for methods on `ActiveServerEntity` and `PassiveServerEntity`.

Example questions:
* Are `connected` and `disconnected` calls serialised or concurrent?
* Are `connected` or `disconnected` calls concurrent or serialised with `invoke`?

I believe this is the kind of information that the Javadoc of the respective types should provide to help entity implementors understand under which guarantees they operate.

jd0-sag:
Posted as a starting-point for discussing these JavaDoc changes.  Once we feel as though this change is complete, we can merge it (potentially requiring some rebase since there are related discussions to start, as well).

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
@@ -16,7 +16,6 @@
  *  Terracotta, Inc., a Software AG company
  *
  */
-
 package org.terracotta.entity;
 
 
@@ -25,23 +24,32 @@ package org.terracotta.entity;
  *  interaction, it has more capabilities than the corresponding passive.
  */
 public interface ActiveServerEntity<M extends EntityMessage, R extends EntityResponse> extends CommonServerEntity<M, R> {
-
   /**
-   * Indicate that the given client is now connected up to this ServerEntity.
+   * <p>Indicate that the given client is now connected up to this ServerEntity.</p>
+   * <p>This is called in response to a client-side fetchEntity() or the reconnect of a client which had previously fetched
+   *  the entity, after a server restart or fail-over.</p>
+   * <p>Note that this call is made on the {@link ConcurrencyStrategy#MANAGEMENT_KEY}, meaning that it is serialized with
+   *  respect to all other messages enqueued for the entity.</p>
    *
    * @param clientDescriptor client-side instance which connected
    */
   void connected(ClientDescriptor clientDescriptor);
 
   /**
-   * Notify the ServerEntity that the given client has disconnected.
+   * <p>Notify the ServerEntity that the given client has disconnected.</p>
+   * <p>This is called in response to a client closing its previously fetched entity or when a client disappears from the
+   *  cluster (due to a network failure, unexpected termination, etc).</p>
+   * <p>Note that this call is made on the {@link ConcurrencyStrategy#MANAGEMENT_KEY}, meaning that it is serialized with
+   *  respect to all other messages enqueued for the entity.</p>
    *
    * @param clientDescriptor client-side instance which disconnected
    */
   void disconnected(ClientDescriptor clientDescriptor);
   
   /**
-   * Invoke a call on the given entity.
+   * <p>Invoke a call on the given entity.</p>
+   * <p>Note that the thread used to make this call is determined by consulting the entity's ConcurrencyStrategy so it may be
+   *  called concurrently with other invokes.</p>
    *
    * @param clientDescriptor source instance from which the invocation originates.
    * @param message The message from the client
@@ -50,9 +58,11 @@ public interface ActiveServerEntity<M extends EntityMessage, R extends EntityRes
   R invoke(ClientDescriptor clientDescriptor, M message);
 
   /**
-   * Called during client reconnect to allow the client to pass arbitrary extra data to the server-side entity so it can
-   * rebuild any in-memory state it had, related to the connected client.
-   * Note that this is called AFTER connected() is called for this clientDescriptor.
+   * <p>Called during client reconnect to allow the client to pass arbitrary extra data to the server-side entity so it can
+   *  rebuild any in-memory state it had, related to the connected client.</p>
+   * <p>Note that this is called AFTER connected() is called for this clientDescriptor.</p>
+   * <p>Note that this call is made on the {@link ConcurrencyStrategy#MANAGEMENT_KEY}, meaning that it is serialized with
+   *  respect to all other messages enqueued for the entity.</p>
    * 
    * @param clientDescriptor The client-side instance which reconnected
    * @param extendedReconnectData Arbitrary data sent by the client-side instance to rebuild the server-side in-memory state 
@@ -60,10 +70,11 @@ public interface ActiveServerEntity<M extends EntityMessage, R extends EntityRes
   void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedReconnectData);
 
   /**
-   * Passes any information required to describe all entity data/state associated with the given concurrency key to a
-   * passive instance being synchronized to be consistent with the receiver.
+   * <p>Passes any information required to describe all entity data/state associated with the given concurrency key to a
+   *  passive instance being synchronized to be consistent with the receiver.</p>
    * 
-   * Note that this method is also run on the concurrencyKey specified, so it blocks other messages executed on that key.
+   * <p>Note that this method is also run on the concurrencyKey specified, so it blocks other messages executed on that
+   *  key.</p>
    * 
    * @param syncChannel The output channel to the passive
    * @param concurrencyKey The key of the data to be synchronized
--- a/entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
@@ -16,46 +16,57 @@
  *  Terracotta, Inc., a Software AG company
  *
  */
-
 package org.terracotta.entity;
 
 
 /**
- * The methods specifically supported by passive entities.  Note that a passive doesn't know anything about
- *  connected clients, nor can it return data to them.
- * Note that passive entities have no invoke response but we still accept the generic R so that accessors of the codec will
- * see a consistent type landscape.
+ * <p>The methods specifically supported by passive entities.  Note that a passive doesn't know anything about
+ *  connected clients, nor can it return data to them.</p>
+ * <p>Note that passive entities have no invoke response but we still accept the generic R so that accessors of the codec
+ *  will see a consistent type landscape.</p>
  */
 public interface PassiveServerEntity<M extends EntityMessage, R extends EntityResponse> extends CommonServerEntity<M, R> {
   /**
-   * Invoke a call on the given entity.  Note that passive entities can't return data to the client.
-   * This method is called both in the cases of normal client->server method invocation but also in the case of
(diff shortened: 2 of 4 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -25,23 +24,32 @@
  *  interaction, it has more capabilities than the corresponding passive.
  */
 public interface ActiveServerEntity<M extends EntityMessage, R extends EntityResponse> extends CommonServerEntity<M, R> {
-
   /**
    * Indicate that the given client is now connected up to this ServerEntity.
+   * This is called in response to a client-side fetchEntity() or the reconnect of a client which had previously fetched the
+   *  entity, after a server restart or fail-over.
+   * Note that this call is made on the MANAGEMENT_KEY, meaning that it is serialized with respect to all other messages
```
ljacomet: I would link to the `MANAGEMENT_KEY` javadoc here and in other comments.

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -25,23 +24,32 @@
  *  interaction, it has more capabilities than the corresponding passive.
  */
 public interface ActiveServerEntity<M extends EntityMessage, R extends EntityResponse> extends CommonServerEntity<M, R> {
-
   /**
    * Indicate that the given client is now connected up to this ServerEntity.
+   * This is called in response to a client-side fetchEntity() or the reconnect of a client which had previously fetched the
```
ljacomet: I would use `<P>` and `</P>` to allow javadoc to render multiple paragraphs.

ljacomet:
Looks good to me - made a few comments on Javadoc format.

AbfrmBlr:
Would it be possible for platform to let entity know about the closure of the reconnect window.
What I understand now is that I can get invokes during the reconnect window. But is there a way for platform to tell that now reconnect window is closed and Entity will not get any `handleReconnect` calls going forward.

jd0-sag:
Updates made.  Take a look and see what you think.

@AbfrmBlr That is a different issue which should be discussed in its own context.

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
```diff
@@ -16,46 +16,56 @@
...
    */
   void endSyncEntity();
 
   /**
-   * Called on concurrencyKey to notify the receiver that it is about to start receiving synchronization messages on this
-   * key.
+   * <p>Called on concurrencyKey to notify the receiver that it is about to start receiving synchronization messages on this
+   *  key.</p>
+   * <p>To clarify:  the concurrencyKey is both the argument (so the entity can logically act on it) but is also the key
+   *  where this call is executed.</p>
```
AbfrmBlr: The call to this methods is done for all concurrency keys returned by `ConcurrencyStrategy.getKeysForSynchronization`.
Is there any guarantee on the order of the keys to be synced. The  cache lifecycle operations are handled by an exclusive concurrency key `1`. Let's say if platform tries to sync other keys before `1`, the sync messages will fall flat on passive in absence of cache. The minimal thing that can work in this case is that either platform let's ehcache to specify a key that should be synced before any other syncing starts. or it allows entity to do initial state transfer during `startSyncEntity`.

jd0-sag: There is no order as the keys are expected to represent largely disjoint data.

I will have to think about the idea of being able to send a sync message before and/or after the rest of the sync, potentially on the MANAGEMENT_KEY since I could imagine these kinds of life-cycle use-cases being a reasonable pattern.

We might have to add something for that, but I will have to think through it.

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -50,20 +58,23 @@
...
   /**
-   * Passes any information required to describe all entity data/state associated with the given concurrency key to a
-   * passive instance being synchronized to be consistent with the receiver.
+   * <p>Passes any information required to describe all entity data/state associated with the given concurrency key to a
+   *  passive instance being synchronized to be consistent with the receiver.</p>
    * 
-   * Note that this method is also run on the concurrencyKey specified, so it blocks other messages executed on that key.
+   * <p>Note that this method is also run on the concurrencyKey specified, so it blocks other messages executed on that
+   *  key.</p>
    *
```
albinsuresh: Do we need to add a note here that the invocation of this method could happen concurrently on different concurrency keys?

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
```diff
@@ -16,46 +16,56 @@
...
 
   /**
-   * Called on concurrencyKey to notify the receiver that synchronization of this key is complete and no more
-   * synchronization messages will be received on that key.
+   * <p>Called on concurrencyKey to notify the receiver that synchronization of this key is complete and no more
+   *  synchronization messages will be received on that key.</p>
+   * <p>To clarify:  the concurrencyKey is both the argument (so the entity can logically act on it) but is also the key
+   *  where this call is executed.</p>
    */
   void endSyncConcurrencyKey(int concurrencyKey);
```
albinsuresh: Shouldn't we add a note here that `invoke` calls on a concurrencyKey will start only after this method has been completed for that same key? And that `invoke` calls on a concurrencyKey does not wait till entire sync is complete (`endSyncEntity(...)`) but can start immediately after the `endSyncConcurrencyKey(...)` invocation is complete for that key?

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
chrisdennis: Since there is no structure available outside of the entities, there is always going to be structure within entities.  That implies that the data will almost never be a flat structure devoid of inter-dependencies.  In the presence of inter-dependencies its going to be very awkward to perform a passive sync without the ability to somehow control the order in which the passive sync data is sent.

jd0-sag: I can think of 2 ways to address that, then:

1. We do run a synchronization on `MANAGEMENT_KEY`, prior to the others (maybe a before and after).
1. We change the `ConcurrencyStrategy.getKeysForSynchronization()` to return a list (it is currently a set) and state that they are always synchronized serially, in the returned order.

Would either of those address this issue?

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
jd0-sag: Yes, but only if we keep that as the current rule.  Some of the other discussion in this review might mean that we change that.  We will need to document that detail once we figure out which way we want it to go.

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
jd0-sag: That is a good idea.  I will add a note about that in my next push (in a moment).

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
myronkscott: Current behavior is to iterate sync in order and serially over the keys as presented by the entity implementor.  This is to minimize the impact of sync on client throughput.  The question is do we want to lock ourselves into this behavior.

ljacomet: See general comment on this point

ljacomet:
@myronkscott and I had a chat on this topic a week ago.

What about transforming the return type of `ConcurrencyStrategy.getKeysForSynchronization` to a type that would allow to express some structure in the way sync needs to happen.

For example, the current Ehcache use case would require one key to be synced first then all others could be synced in parallel (and thus any order).

So to make this more generic, I would see the return be something containing of a list of keys and a set of keys. With the contract that the list has to be processed in order and serially while the set content can be processed in parallel.

Wdyt? @chrisdennis Anything you can think of that would require such a scheme to be even more general? recursive maybe? more complex?

ChrisGreenaway:
It seems like a more general version of what Louis suggests would be to return a list of sets of keys.

The sets must be processed in the order specified by the outer list but, within each set, the order doesn't matter.

ljacomet:
@ChrisGreenaway Correct - that's what came up after discussing this topic with others yesterday.

So we would move to `List<Set<Integer>> getKeysForSynchronization()`.

@jd0-sag Can you add this in this PR or do you want me to open a separate one?

jd0-sag:
The concern that we have with this is how to handle the case of the same key being specified multiple times, in different sets. (since that would be **very** bad)

We are sort of wondering if we should just document this to state that the set will be iterated in its definition of "order" so that an entity which wants keys synchronized in a certain order should just use some kind of ordered set.

ljacomet:
I understand the concern. But it would break the entity and nothing else, right?

As for sticking with a single level `Set`, it seems to me that future optimisation - whether they effectively happen or not - are suddenly ruled out. And this feels bad, especially if it only is driven by _protecting_ the entity implementor.

One more question to this discussion: when passive sync starts and you process the first concurrency key, in which state are the others? let's say I have concurrency keys 1, 2, 3 and 4 to be processed in that order (to simplify the discussion). When processing 1, are 2,3 and 4 blocked or running? When processing 3, are 1, 2 and 4 blocked or running?

ChrisGreenaway:
I see several different solutions to this. Here are a couple:

1) There's no reason to make the getKeysForSynchronization() return type explicitly a List<Set<Integer>>. It could be a KeySyncSpec, which internally has a List<Set<Integer>> and enforces the contract that there are no duplicated integers.

2) The platform could check that there are no duplicates in the List<Set<Integer>>. This is only an O(N) operation. Often an entity is returning the same object every time (right?), so the platform need only check the first time.

jd0-sag:
Given that we have typically viewed the entity API as something we can change, in the future, do we want to over-design it in this particular way?  Shouldn't we just reflect the simplicity of the approach and worry about other approaches (which may be completely different) when/if we go down that road?

Part of the issue with validating the key set at runtime is that there is no good way of handling "you specified a strategy which overlaps itself".  It opens a door we can't reasonably handle in support a feature we don't have.

Regarding _other_ keys while one is syncing - only the key being synced is blocked.  This means that an entity can have a partially synced state so we don't stop everything across the cluster to wait for it to finish.

ljacomet:
@jd0-sag Regarding the syncing - that's interesting. In the context of the discussions about having dynamic concurrency key scopes for syncing, Ehcache would need to have the guarantee that between the request of the keys to sync and the syncing of the first of these, no operations happen.
Is that something which can be provided?

myronkscott:
A complete rewrite on the way passive sync works would need to happen in order to take advantage of the extra information a "List of Sets" provides.  This feels a bit like a premature optimization and I'm not sure parallel sync of keys would actually be better.  If we go that route, I think we should version the API when the passive sync rewrite happens.

jd0-sag:
Updated this to describe the keys as being synchronized in-order, one at a time.

PR #147, opened 2016-09-06
Fix #145: Expanded documentation surrounding passive sync

### LINKED ISSUE #145: ServerEntity API use of ConcurrencyStrategy UNIVERSAL_KEY
Please add:
1) a ConcurrencyStrategy that returns UNIVERSAL_KEY for all messages (in the same vein as NoConcurrencyStrategy)
2) documentation to ConcurrencyStrategy.getKeysForSynchronization() to explain what set of keys should be used when concurrencyKey can return UNIVERSAL_KEY or MANAGEMENT_KEY.

jd0-sag:
Posted as a starting-point for discussing these JavaDoc changes. Once we feel as though this change is complete, we can merge it (potentially requiring some rebase since there are related discussions to start, as well).

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
@@ -22,40 +22,51 @@ import java.util.Set;
 
 
 /**
- * The concurrency strategy is used to provide the concurrent execution hints to the platform, for a given EntityMessage.
+ * <p>The concurrency strategy is used to provide the concurrent execution hints to the platform, for a given
+ *  {@link EntityMessage}.</p>
  * 
- * NOTE:  A given EntityMessage can only have one concurrency key and the same key must be returned each time the receiver
+ * <p>NOTE:  A given EntityMessage can only have one concurrency key and the same key must be returned each time the receiver
  * is asked the key of a specific message.  There are no constraints between messages but a specific message must always
- * resolve to the same key.
+ * resolve to the same key.</p>
+ * <p>Additionally, the namespace of concurrency keys is defined on a per-entity basis.  Entities are already allowed to be
+ * scheduled independently.  This may have implications for interactions with services as this is the only place where
+ * entities can interact with the same component (the {@link ServiceProvider}).</p>
  * 
  * @param <M> The message type expected by the implementation.
  */
 public interface ConcurrencyStrategy<M extends EntityMessage> {
   /**
-   * UNIVERSAL_KEY is a negative key that indicates a request requires no order and can be run concurrently with any
-   * other entity operation.
+   * <p>UNIVERSAL_KEY is a negative key that indicates a request requires no order and can be run concurrently with any
+   * other entity operation.</p>
    */
   public final int UNIVERSAL_KEY = Integer.MIN_VALUE; 
   /**
-   * MANAGEMENT_KEY is a zero key which is used to manage the entity (anything other than invoke action).
+   * <p>MANAGEMENT_KEY is a zero key which is used to manage the entity (anything other than invoke action).</p>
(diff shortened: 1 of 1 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
```diff
@@ -38,20 +41,22 @@
   public final int UNIVERSAL_KEY = Integer.MIN_VALUE; 
   /**
    * MANAGEMENT_KEY is a zero key which is used to manage the entity (anything other than invoke action).
+   * Additionally, a message can choose to use this key if it must be run exclusively within the entity.  Be careful with
+   * the cases where this is done, however, since the MANAGEMENT_KEY cannot be used for synchronizing data.
```
ChrisGreenaway: "the MANAGEMENT_KEY cannot be used for synchronizing data" - can you explain what you mean by this please?

entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
```diff
@@ -67,6 +72,15 @@
    * server.  This operation is performed on a per-concurrencyKey basis so this method is used to determine the set of keys
    * which must be synchronized.
    * 
+   * A synchronization operation will be issued for each of the keys returned and run on that key.  This means that the
```
ChrisGreenaway: When you say "a synchronization operation" - do you mean a call to synchronizeKeyToPassive on the active entity. If so, say that, if not, then say what.

entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
myronkscott: I wonder if MANAGEMENT_KEY actually has much meaning for an entity implementation.  For the platform, it's used for operations that are fully exclusive for an entity like lifecycle operations where no other operations can be running on any concurrency key.  If an entity needs to have 100% serial execution it makes more sense for that entity to define one concurrency > 0 and use that key exclusively.  Therefore, I'm having a hard time coming up with a scenario where an entity should actually use the MANAGEMENT_KEY in an invoke.

jd0-sag:
Pushed some updates so let me know what you think.

ljacomet:
* On the topic: Some wording may change due to the discussions in #146 
* On the presentation: I would add html paragraphs separators where the javadoc text uses line breaks. I believe this improves javadoc readability instead of having a single paragraph when rendered.

entity-server-api/src/main/java/org/terracotta/entity/ConcurrencyStrategy.java
```diff
@@ -38,20 +41,24 @@
   public final int UNIVERSAL_KEY = Integer.MIN_VALUE;
```
ljacomet: From discussion with @myronkscott looks like this key also has the property of trying to pick the shortest path. That would be a great property for Ehcache read operations for example. Can this be confirmed?

jd0-sag: Depends on what you mean by "shortest path".  While I am not sure how aggressive the current implementation is, the plan is that any message scheduled with this key will be enqueued in the shortest execution queue.  This is possible since it provides no ordering guarantees.

Whether or not this is a good idea for EhCache will depend entirely on the way per-key data is stored and accessed.  If concurrent access is possible, then this might be worth looking at.  If the storage is optimized for single-threaded access, then you still need to schedule it in the appropriate key.

We probably want to get collect some data regarding the execution queue depth before trying to use this optimization in a situation where there would be any trade-off.  If the queues stay short or are evenly distributed (in a typical use-case), then this wouldn't do much to help.

jd0-sag:
This can be merged (unless any other changes require it to rebase, first).

PR #148, opened 2016-09-06
Fix #138: Clarify the implications of using something from one service in another

### LINKED ISSUE #138: Interaction between services
While developing state service, there was a need for services to interact among themselves. As of now, there is nothing in api which allows this. But it can be done by passing the `ServiceRegistry` reference at the entity level to the services using the `ServiceConfiguration`. As of now this works, since the api is not explicit about the interaction, will trying to get service using the service registry passed on to the entities be a problem. Or do we need an explicit boundaries defined about service registry usage.

jd0-sag:
-this also goes into some detail of how consumerID is used to allow a ServiceProvider to implement a basic namespacing of services but that this is the only implicit namespacing mechanism provided, meaning that an entity uses the same consumerID, no matter how many times it requests the service or who it shares the returned instance with

Posted as a starting-point for discussing these JavaDoc changes. Once we feel as though this change is complete, we can merge it (potentially requiring some rebase since there are related discussions to start, as well).

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
@@ -16,24 +16,39 @@
  *  Terracotta, Inc., a Software AG company
  *
  */
-
 package org.terracotta.entity;
 
+
 /**
- * Registry which Entities interact with list of services provided by the platform. Services provided to entity might
- * be namespaced or not, depending on  service provider.
- *
- * @author twu
+ * <p>The registry given to entities so that they can look up the services hosted on the platform.</p>
+ * 
+ * <p>Note that the Service instances returned by this interface are not cached or interned, in any way (unless the
+ *  underlying ServiceProvider decides to add that functionality).  This means that multiple requests for the same service,
+ *  through the same registry, could return different instances.</p>
+ * 
+ * <p>ServiceProvider implementations see each call through a ServiceRegistry as originating from a specific consumerID,
+ *  which is per-entity.  Using this mechanism is how a service knows which entity is contacting it, no matter how many
+ *  instances of its underlying service were returned through repeated getService calls.  More information regarding
+ *  consumerIDs can be found in {@link ServiceProvider}.</p>
+ * 
+ * <p>Note about using services together:  Services are not explicitly granted access to each other as there would be no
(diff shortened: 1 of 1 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
```diff
@@ -16,24 +16,38 @@
...
 /**
- * Registry which Entities interact with list of services provided by the platform. Services provided to entity might
- * be namespaced or not, depending on  service provider.
- *
- * @author twu
+ * The registry given to entities so that they can look up the services hosted on the platform.
+ * 
+ * Note that the Service instances returned by this interface are not cached or interned, in any way (unless the underlying
+ *  ServiceProvider decides to add that functionality).  This means that multiple requests for the same service, through the
+ *  same registry, will return different instances.
```
ChrisGreenaway: "will return different instances" is presumably not true if "the underlying ServiceProvider decides to add that functionality".

entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
```diff
@@ -16,24 +16,38 @@
...
+ * 
+ * ServiceProvider implementations see each call through a ServiceRegistry as originating from a specific consumerID, which
```
ChrisGreenaway: Does this mean that an entity operating within one concurrency key will have the same consumerID as the same entity operating within a different concurrency key? Does that matter? Are consumerIDs used like concurrencyIDs to ensure thread safety in the service? If not, then why is it only possible to use services together when "the entity and calling service know that they aren't going to collide"?

entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
ljacomet: Indeed - might be better to use _could_ instead of _will_

entity-server-api/src/main/java/org/terracotta/entity/ServiceRegistry.java
jd0-sag: consumerIDs are described in greater detail in ServiceProvider so I mentioned that as a link in this update.

jd0-sag:
Updated in response to comments.

jd0-sag:
This can be merged (unless any other changes require it to rebase, first).

PR #151, opened 2016-09-12
(DISCUSS) Create IStripeMonitoring interface for receiving monitoring data from a stripe

jd0-sag:
* this splits the way that monitoring data moves through the system into explicit "sender" and "receiver" components
* `IMonitoringProducer` is still used by an entity when it wants to send data to the monitoring implementation
* `IStripeMonitoring` is the new interface which must be implemented by the monitoring component in order to receive this data
* the `IStripeMonitoring` implementation will receive all monitoring data from the entire stripe, but only the instance running on the active will be given this information

### DIFF
--- a/monitoring-support/src/main/java/org/terracotta/monitoring/PlatformMonitoringConstants.java
+++ b/monitoring-support/src/main/java/org/terracotta/monitoring/PlatformMonitoringConstants.java
@@ -33,10 +33,6 @@ public class PlatformMonitoringConstants {
    */
   public static final String PLATFORM_ROOT_NAME = "platform";
   /**
-   * The name of the node in the tree which is the parent to all nodes representing connected servers.
-   */
-  public static final String SERVERS_ROOT_NAME = "servers";
-  /**
    * The name of the node in the tree which is the parent to all nodes representing connected clients.
    */
   public static final String CLIENTS_ROOT_NAME = "clients";
@@ -49,7 +45,8 @@ public class PlatformMonitoringConstants {
    */
   public static final String FETCHED_ROOT_NAME = "fetched";
   /**
-   * The name of the node in the tree which stores the name of the server's current state.
+   * The name of the node in the tree which stores the ServerState instance.
+   * This is a direct child of PLATFORM_ROOT_NAME.
    */
   public static final String STATE_NODE_NAME = "state";
 
@@ -57,10 +54,6 @@ public class PlatformMonitoringConstants {
    * The path of the platform node, for manipulating its children.
    */
   public static final String[] PLATFORM_PATH = {PLATFORM_ROOT_NAME};
-  /**
-   * The path of the platform's clients node, for manipulating its children.
-   */
-  public static final String[] SERVERS_PATH = {PLATFORM_ROOT_NAME, SERVERS_ROOT_NAME};
   /**
    * The path of the platform's clients node, for manipulating its children.
    */
--- a/standard-cluster-services/pom.xml
+++ b/standard-cluster-services/pom.xml
@@ -42,6 +42,11 @@
       <artifactId>packaging-support</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta</groupId>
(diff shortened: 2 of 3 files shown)


jd0-sag:
Note that one of the other things I need to change in this PR (which I will probably make as a different commit, but in the same PR) is removing `PlatformMonitoringConstants.SERVERS_PATH`.  This doesn't really make sense if we are explicitly handling server data as parallel trees so I will probably replace this with specific paths for the server data we wanted to store.

mathieucarbou:
@jd0-sag : ok let us know. I would need some more info about `IStripeMonitoring `: do you mean that the monitoring service implementation will need to implement `IStripeMonitoring `, or that the monitoring service must also supply a service of type `IStripeMonitoring `, that will be responsible of receiving all the entire stripe data ?

Also, I don't really get the real difference between `IStripeMonitoring` and `iMonitoringProducer`. Especially: if `IStripeMonitoring` is "provided as a service", what prevents en entity to request a `IStripeMonitoring` and do some calls ?

jd0-sag:
This means that the service you would be providing needs to implement `IStripeMonitoring` _instead_ of `IMonitoringProducer`.

While nothing would prevent an entity from reaching into `IStripeMonitoring`, directly, it wouldn't make sense to call it directly since the entity doesn't have the server identifier.  They should continue to use `IMonitoringProducer`, as they currently do.

The simplest way to think of it is that the platform places a shim between these 2 interface to provide the extra per-server identification for each call passing through (that is literally what it does on the active while the passive does that but also involves the network).

mathieucarbou:
ok!

About `PlatformMonitoringConstants.SERVERS_PATH`: I see our code heavily depends on that, and the fact that `addNode(PlatformMonitoringConstants.SERVERS_PATH, PlatformServer)` is called.

As I understand know, there won't be any `SERVERS_PATH` ? So...

- what would be the paths used to expose server information (and especially the state) ?
- what interface will be used to expose server information and by who ? I.e. in our implementation, we listen for some platform paths to generate events. So Do we need to listen only in `IStripeMonitoring ` ? What I mean is: do `IMonitoringProducer` is a simple delegate to a `IStripeMonitoring ` on teh smae stripe ? Do we only need to implement `IStripeMonitoring` and when providng a `IMonitoringProducer`, we return an implementation working on the right server and delegating to a `IStripeMonitoring` ?

jd0-sag:
My plan is to replace `PlatformMonitoringConstants.SERVERS_PATH` with 2 new path constants:  one for the `PlatformServer` and one for the `ServerState`.

You will _only_ be producing an implementation of `IStripeMonitoring`, no longer providing anything related to `IMonitoringProducer` since it will now just be this internal shim.  It will work the same way, except that there is an additional argument for the sending server.

jd0-sag:
I added the commit to change the path constants.

jd0-sag:
Some things from discussions today:

* Given that we already need some kind of token to identify a server, we should probably use `PlatformServer`, directly, since that is already defined as part of the tc-apis for monitoring
* If we want to use `PlatformServer`, we will also need some basic life-cycle methods so we know when each one is valid and what it means, at a high level (since this information otherwise requires sniffing through the platform's monitoring tree paths):  `serverDidBecomeActive()` (called once, before any other methods, when the server first becomes active), `serverDidJoinStripe()` (called whenever another server joins the stripe - not the active), `serverDidLeaveStripe()` (called whenever a server leaves the stripe)
* This means that we can remove the now-redundant `PlatformServer` instance from the monitoring tree

In the future, we may want to create other methods to replace the need for `PlatformMonitoringConstants`, entirely.  Given that this new interface is specifically intended to be called by the active server, only, it seems reasonably appropriate.  Whether that is done at all, done as a change to this interface, or done as an entirely different one, is a different matter to discuss in the future.

jd0-sag:
I have a prototype of this change working so I updated the commits in the PR.

Of specific note is that the life-cycle calls MUST be only ever delivered on consumerID 0.  This is because the `ServiceProvider` doesn't expose an interface, only its provided services do.  These are always per-consumerID, so we need to define a specific receiver for this "special" calls.

PR #154, opened 2016-09-20
[DO NOT MERGE] Message typing for Replication

myronkscott:
This is for discussion only.  We should probably pick one method or the other but not both.

One method uses the type system to designate if the message is active or passive only.  The common type means run on both active and passive

```
class MyActiveMessage extends ActiveOnlyEntityMessage {  //  active only execution

}

class MyPassiveMessage extends PassiveOnlyEntityMessage {  //  passive only execution

}

class MyEnytityMessage extends EntityMessage {  // runs on both 

}
```

The other method uses annotation to convey the same information

```
@Execution(Execution.ServerState.ACTIVE)
class MyActiveMessage extends EntityMessage {  // active only 

}

@Execution(Exection.ServerState.PASSIVE) 
class MyPassiveMessage extends EntityMessage {  // passive only 

}

@Execution({Exection.ServerState.ACTIVE, Exection.ServerState.PASSIVE})
class MyEntityMessage extends EntityMessage {  //  both

}

class MyEntityMessage {  // both

}
```

### DIFF
--- a/entity-common-api/src/main/java/org/terracotta/entity/ActiveExcludedEntityMessage.java
+++ b/entity-common-api/src/main/java/org/terracotta/entity/ActiveExcludedEntityMessage.java
@@ -1,0 +1,24 @@
+/*
+ *
+ *  The contents of this file are subject to the Terracotta Public License Version
+ *  2.0 (the "License"); You may not use this file except in compliance with the
+ *  License. You may obtain a copy of the License at
+ *
+ *  http://terracotta.org/legal/terracotta-public-license.
+ *
+ *  Software distributed under the License is distributed on an "AS IS" basis,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ *  the specific language governing rights and limitations under the License.
+ *
+ *  The Covered Software is Entity API.
+ *
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+
+package org.terracotta.entity;
+
+
+public interface ActiveExcludedEntityMessage extends EntityMessage {
+}
--- a/entity-common-api/src/main/java/org/terracotta/entity/Execution.java
+++ b/entity-common-api/src/main/java/org/terracotta/entity/Execution.java
@@ -1,0 +1,33 @@
+/*
+ *
+ *  The contents of this file are subject to the Terracotta Public License Version
+ *  2.0 (the "License"); You may not use this file except in compliance with the
+ *  License. You may obtain a copy of the License at
+ *
+ *  http://terracotta.org/legal/terracotta-public-license.
+ *
+ *  Software distributed under the License is distributed on an "AS IS" basis,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ *  the specific language governing rights and limitations under the License.
+ *
+ *  The Covered Software is Terracotta Core.
+ *
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+package org.terracotta.entity;
+
+import java.lang.annotation.ElementType;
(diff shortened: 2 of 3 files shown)


chrisdennis:
Regarding the proposal `*Only` semantics don't make sense when multiple interface inheritance is possible.  So regardless of which solution we use the effects have to be additive.  How sure are we that messages.  If it is to be done in the type system I'm generally pro marker interfaces because they can't silently fail, and they form a proper part of the type system (i.e. you can enforce where they get to be passed/consumed).

In a wider sense I'm not 100% on whether putting this in the hands of a type system is the right way to do it...

myronkscott:
We could use method calls but that will break everyone since default interfaces are Java8 only.  I agree that `class MyMessage implements ActiveOnly, PassiveOnly, EntityMessage {...}` is bad but I don't think additive will work then since the base class will have no meaning.  Is there a way forward?

myronkscott:
Would it be sufficient that implementation Asserts that implementing both interfaces is nonsensical?  

The problem here is that setting replicate flag on the InvocationBuilder is misplaced as it is possible to send the exact same message to the cluster as replicated and non-replicated which in all cases seems like a bug.

One other option is to remove. `InvocationBuilder.replicate()` and just replicate everything, requiring entity implementors to ignore messages that are inappropriate for active or passive context.

myronkscott:
How about:
```
class MyActiveMessage implements PassiveExcludedEntityMessage { // runs on active only

}

class MyPassiveMessage implements ActiveExcludedEntityMessage { // runs on passive only

}

class MyNoneMessage implements ActiveExcludedEntityMessage , PassiveExcludedEntityMessage { // runs nowhere

}

class MyEverywhereMessage implements EntityMessage {  // runs on both

}
```

myronkscott:
Some other options might be to add an additional interface that can be optionally added to the interface that tells the platform where the message can run.

ljacomet:
I really do not like the _negation_ in name approach. It is always hard to read.
And the remark from @chrisdennis about not being sure if that information really needs to be in the type system gets me thinking.

What about adding:
`MessageMode EntityMessage.getMessageMode()`
where `MessageMode` would have the values:
* `ACTIVE`
* `PASSIVE`
* `BOTH`
(names to be improved - let's discuss the concept first)

This has the added benefit that a given message type could behave differently. For example, an active needing to send a message to the passive after receiving it from the client. Message is the same, just different payload -> single message type, with flag being different.

I understand it means breaking all existing `EntityMessage` implementations, but since we control when these changes land in the different repositories making use of the API, I do not believe it is such a big problem.

myronkscott:
OK.  Looks like we are moving away from the type system but I'm not sure I like a method call.  Part of the reason that I chose the type system is that I wanted compile time designation of execution location by the entity implementor.  The other issue is that it would be nice if EntityMessage stayed "just data".  Adding methods to it could start giving it behavior and I'm not sure that is what is desired.  How about adding `ExecutionStrategy` to the ServerEntityService.  Much like `ConcurrencyStrategy`, this interface could be consulted on where execution is desired for a particular message.

myronkscott:
Circling back to the main issue that started this line of thought.  I think `InvocationBuilder.message().replicate().invoke()` is a bad pattern.  It implies that 
```
InvocationBuilder.message(new MutationMessage()).replicate(true).invoke();
InvocationBuilder.message(new MutationMessage()).replicate(false).invoke();
```
is OK.  In every case I can think of, this would be a bug.  If we do not feel this needs adjustment, we can always leave it how it currently stands.

ljacomet:
I agree this needs to be changed, because as you said the pattern for most cases should be a bug.
The reason I thought about having this on the message is that it enables an entity implementor to have the same message type to be used either between client and active or active and passive for example. And in this case, the target of the message feels like it is data and thus belongs on it.

Maybe the `ExecutionStrategy` allows entity implementors to choose between a static type based system and a message property indeed. While the method on `EntityMessage` only allows the later.

PR #163, opened 2016-10-14
API changes entities

myronkscott:
Added reconfigure to EntityServerService so that implementors have an opportunity to transfer state of the entity to new reconfigured entity since reconfigure happens in the middle of entity execution.

Added ExecutionStrategy to deprecate `InvocationBuilder.requiresReplication()` since that call is mis-placed and mis-used by the platform.

Updated the pom files to reflect that the server is java 1.8 only.

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/EntityServerService.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/EntityServerService.java
@@ -64,6 +64,30 @@ public interface EntityServerService<M extends EntityMessage, R extends EntityRe
   PassiveServerEntity<M, R> createPassiveEntity(ServiceRegistry registry, byte[] configuration);
   
   /**
+   * Reconfigure an existing entity during server execution.  Reconfigure call in the
+   * platform's entity API allows an entity to reconfigure in place, its instance in the 
+   * system.  Prior to the call of this method, the entities execution queue is flushed and 
+   * any subsequent messages are blocked until this call completes.  This call occurs only 
+   * under the MANAGEMENT_KEY concurrency key.  Implementations are allowed to either reconfigure
+   * in place and return the same instance of the entity or a new entity may be built if desired.
+   * 
+   * @param <AP> the entity type being reconfigured
+   * @param registry the service registry for the entity
+   * @param oldEntity the instance of the entity implementation being reconfigured
+   * @param configuration the new configuration of the entity
+   * @return An entity instance influenced by the new configuration
+   */
+  default <AP extends CommonServerEntity<M, R>> AP reconfigureEntity(ServiceRegistry registry, AP oldEntity, byte[] configuration) {
+    if (oldEntity instanceof PassiveServerEntity) {
+      return (AP)createPassiveEntity(registry, configuration);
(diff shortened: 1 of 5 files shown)


jd0-sag:
This change regarding the `ClientDescriptor` confuses me.  I will probably need some nuance explained to me (I assume it is something regarding the default method uses).  Other than that, the commit fixing the compilation issues should be squashed into the commits which introduced that problem so we don't have an unstable hole in the version history.

entity-server-api/src/main/java/org/terracotta/entity/EntityServerService.java
```diff
@@ -64,12 +64,51 @@
...
+   * under the MANAGEMENT_KEY concurrency key.  Implementations are allowed to either reconfigure
+   * in place and return the same instance of the entity or a new entity may be built if desired.
+   * 
+   * @param <AP> the entity type being reconfigured
+   * @param registry the service registry for the entity
+   * @param oldEntity the instance of the entity implementation being reconfigured
+   * @param configuration the new configuration of the entity
+   * @return An entity instance influenced by the new configuration
+   */
+  default <AP extends CommonServerEntity<M, R>> AP reconfigureEntity(ServiceRegistry registry, AP oldEntity, byte[] configuration) {
```
ljacomet: Surprised that this is considered ok while the discussions around passive entity becoming active could not happen with such a hand over.
What is the fundamental difference between the two?

myronkscott: This does not involve a major state transition of the server.  Connected clients and the invoke queue can be assured to be consistent during the reconfigure call.  With failover, we are dealing with a major state transition of the server where the availability of clients can change during transition.  

Having said that, I am not 100% comfortable with this addition to the API.  I think an option may be to remove reconfigure and force removal and re-creation of entities to produce a similar result.  Coupled with some of the discussions around higher level entity management, this may be a better solution.

myronkscott: Also, restart and failover consistency would be broken if we added such a call for failover.

monitoring-support/src/main/java/org/terracotta/monitoring/PlatformClientFetchedEntity.java
```diff
@@ -36,13 +35,13 @@
 
   public String clientIdentifier;
   public String entityIdentifier;
-  public transient ClientDescriptor clientDescriptor;
+  public transient Object clientDescriptor;
```
myronkscott: @mathieucarbou : can you check to make sure you don't use the client ClientDescriptor here by direct access?

mathieucarbou: Yes we do rely on this field by direct access, and, moreover, we rely on its type also. Why this change ?

myronkscott: Java6/8 madness.  I will go with a more invasive change.  Confirm please that you are using these objects only server-side and java8?

PR #170, opened 2016-10-30
Fix #169: Create new IPlatformPersistence to replace IPersistentStorage

### LINKED ISSUE #169: Introduce API for new platform persistence service
`IPersistentStorage` has many implicit assumptions regarding how it will be implemented which are riven through much of its interface and required usage patterns.  Additionally, much of the API isn't actually used by the platform (which was the primary user, driving initial creation of this service).  These make implementations time-consuming to build, difficult to test, and nearly impossible to optimize.  We need a simpler persistence interface for the platform which can lend itself, more easily, to higher-performance implementations.

jd0-sag:
-this new interface is much smaller and more directly aimed at providing high-performance solutions to the platform's persistence needs
-IPersistentStorage and KeyValueStorage can be removed, in a future release, as they are now deprecated


jd0-sag:
Will self-merge, as previously discussed.

PR #172, opened 2016-11-07
(READY) Fix #158: Add delayed/periodic calls to IEntityMessenger

### LINKED ISSUE #158: Task scheduling and async task execution support
For stat computation to work, we need from a server entity to be able to pass a `ScheduledExecutorService` to the stat library.

Since Voltron does not allow to spawn threads, we need a way to support scheduling in Voltron so that we can provide an adapted implementation of a ScheduledExecutorService to this stat library.

Tasks registered in it would have their execution triggered thanks to a platform callback somewhere at a specified interval.

jd0-sag:
-these new interfaces will allow an entity to send itself a specific message after a delay or periodically
-a ScheduledToken is provided so that these calls can be cancelled

NOTE:  This can be merged once we are ready to release.

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
@@ -47,4 +47,45 @@ public interface IEntityMessenger {
    * @throws MessageCodecException The message could not be serialized.
    */
   void messageSelfAndDeferRetirement(EntityMessage originalMessageToDefer, EntityMessage newMessageToSchedule) throws MessageCodecException;
+
+  /**
+   * Requests that the given message be send to the entity instance after millisBeforeSend have elapsed.
+   * The message will be sent once, and may arrive late but won't arrive before millisBeforeSend have elapsed.
+   * The message will not arrive if the server restarts or the entity is destroyed before it was due.
+   * 
+   * @param message The new message to send.
+   * @param millisBeforeSend The minimum number of milliseconds which must pass before message should be delivered.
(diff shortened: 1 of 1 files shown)


myronkscott:
Do you think we should add a 3 arg version that has a delay and period?

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
```diff
@@ -47,4 +47,45 @@
...
+   * Requests that the given message be send to the entity instance after millisBeforeSend have elapsed.
+   * The message will be sent once, and may arrive late but won't arrive before millisBeforeSend have elapsed.
+   * The message will not arrive if the server restarts or the entity is destroyed before it was due.
+   * 
+   * @param message The new message to send.
+   * @param millisBeforeSend The minimum number of milliseconds which must pass before message should be delivered.
+   * @return A token to describe this delayed send which can be used to cancel it.
+   * @throws MessageCodecException The message could not be serialized.
+   */
+  ScheduledToken messageSelfAfterDelay(EntityMessage message, long millisBeforeSend) throws MessageCodecException;
```
myronkscott: Wondering if this should be a Supplier<EntityMessage> so that a new version can be created on each iteration

jd0-sag:
If there is a common use-case to have that, in the future, we can add it.  I don't want to too heavily over-design this, in this first implementation.  The 2 scheduling mechanisms represent a fairly minimal set of primitives and that compound use-case could even be implemented on top of them, if it were a one-off case.

PR #174, opened 2016-11-10
Add basic stripe information querying to the platform service

myronkscott:
There is a need to query very basic information about the stripe from a common service.


myronkscott:
Closed.  Provides not value at the moment.  May revisit at a later point.

PR #193, opened 2017-01-13
Fix #192 - Expose platform configuration file

### LINKED ISSUE #192: Add the ability to grab stripe configuration
PlatformServices should expose the tc-config the server was started with.

myronkscott:
expose the raw platform configuration file to PlatformService

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
@@ -19,6 +19,7 @@
 package org.terracotta.monitoring;
 
 import com.tc.classloader.CommonComponent;
+import java.io.InputStream;
 
 
 /**
@@ -30,4 +31,8 @@ public interface PlatformService {
    * Dumps platform state into logs
    */
   void dumpPlatformState();
+  /**
+   * Provides the raw config file from server startup
+   */
+  InputStream getPlatformConfiguration();
 }


standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -30,4 +31,8 @@
    * Dumps platform state into logs
    */
   void dumpPlatformState();
+  /**
+   * Provides the raw config file from server startup
+   */
+  InputStream getPlatformConfiguration();
```
ljacomet: Can't we make that a String that contains the content of the config file?

myronkscott: I'd rather have the downstream consumer do the conversion.  We can do it prior but that just means possibility messing up the character set.  I believe some consumers (jaxb) have the ability to detect things like that if the file is not in the platforms default. That being said, I'm not opposed to doing the conversion to the os default if that is what is truly desired.

ljacomet: Had not thought of the charset problem. Let's go with that, it is an easy thing to evolve should we decide to.

anthonydahanne: UTF-8 only ?

myronkscott: Can you clarify?  Are you saying that we should mandate config file writers only use UTF-8?

PR #201, opened 2017-02-06
Fix #200:  Flesh out the details of the acknowledgement system in JavaDoc

### LINKED ISSUE #200: Provide greater detail and clarity in message acknowledgement documentation
The acknowledgements requested to `InvocationBuilder` are key to interacting with the cluster but are not described in great detail.  Additionally, much of the current documentation refers to older use-models, not as relevant in modern applications of the platform.

jd0-sag:
(waiting for feedback before merging)


jd0-sag:
Updated title since there doesn't seem to be any concern with this direction.

PR #211, opened 2017-04-03
Fix #202: Added userData argument to EntityClientService.create, from fetchEntity()

### LINKED ISSUE #202: Allow extra "user data" argument through EntityRef.fetchEntity to EntityClientService.create
There is currently no way to pass additional data through to the constructor of a client-side entity (endpoint wrapper).  Allowing this would make it possible to write smaller and more restrictive code (via final variables and other immutability concepts) within the client-side entity.

This would require changing some parts of the API:

At the top-level, this would require changing `EntityRef.fetchEntity()` into `EntityRef.fetchEntity(Object)`.

At the level of the user-provided `EntityClientService`, it would change the interface from `EntityClientService.create(EntityClientEndpoint)` to `EntityClientService.create(EntityClientEndpoint, Object)`.

Note that we should be able to make these into real generics instead of just `Object`.

jd0-sag:
DO NOT MERGE:  Currently posted to check that this makes sense to everyone - release will be required, after merge, to unblock downstream PRs which depend on this.

### DIFF
--- a/connection-api/src/main/java/org/terracotta/connection/Connection.java
+++ b/connection-api/src/main/java/org/terracotta/connection/Connection.java
@@ -38,8 +38,9 @@ public interface Connection extends Closeable {
    * @param name name of the entity
    * @param <T> entity type
    * @param <C> configuration type
+   * @param <U> user data type for EntityClientService
    * @return reference to the entity
    * @throws EntityNotProvidedException There is no client-side service providing entities of type T
    */
-  <T extends Entity, C> EntityRef<T, C> getEntityRef(Class<T> cls, long version, String name) throws EntityNotProvidedException;
+  <T extends Entity, C, U> EntityRef<T, C, U> getEntityRef(Class<T> cls, long version, String name) throws EntityNotProvidedException;
 }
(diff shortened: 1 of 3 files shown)


jd0-sag:
@ChrisGreenaway Let me know if this is consistent with your expectations.

ChrisGreenaway:
Looks reasonable overall. I wonder whether it would be worth keeping a variant of fetchEntity() that doesn't take any user data.

jd0-sag:
Would have been worth looking at if it would avoid breakage but the change to the `EntityClientService` interface means that there will already be breakage.  Hence, we might as well avoid the overloading.

ChrisGreenaway:
Not to avoid breakage, but to cope with the situation where an entity doesn't have any user data - otherwise what do you do - pass null? That's not very nice.

jd0-sag:
Passing null is explicit and easy.  Overloading is hidden.

ChrisGreenaway:
I guess we just have different opinions on the niceness of passing nulls and how hidden overloading is.

jd0-sag:
Yeah.  I would be more interested in leaving something else in if would substantially reduce downstream breakage but the change of the interface method combined with the introduction of the type parameter on the EntityRef means that this will already hit both the implementations and the callers.

PR #214, opened 2017-04-20
Issue 510: Retirement Handles for explicit retirement

cschanck:
Add retirement handles and explicit deferment/retirement to
IEntityMessenger. This is done by deferring retirement of one
message based on another, but not delivering the future message
yet. The returned handle will deliver the future message on it's
release.

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
@@ -1,0 +1,28 @@
+/*
+ *
+ *  The contents of this file are subject to the Terracotta Public License Version
+ *  2.0 (the "License"); You may not use this file except in compliance with the
+ *  License. You may obtain a copy of the License at
+ *
+ *  http://terracotta.org/legal/terracotta-public-license.
+ *
+ *  Software distributed under the License is distributed on an "AS IS" basis,
+ *  WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ *  the specific language governing rights and limitations under the License.
+ *
+ *  The Covered Software is Entity API.
+ *
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+package org.terracotta.entity;
+
+import java.util.concurrent.TimeUnit;
+
+public interface ExplicitRetirementHandle {
+  String getTag();
+
+  void release() throws MessageCodecException;
+
+}
--- a/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
@@ -39,6 +39,18 @@ public interface IEntityMessenger {
   void messageSelf(EntityMessage message) throws MessageCodecException;
 
   /**
+   * Defer a message based on another message, which will be sent in the future.
+   *
+   * @param tag String reason tag, mainly for debugging.
+   * @param originalMessageToDefer original message to defer
+   * @param futureMessage message to use to complete original message
+   * @return handle used to release the deferred message
+   */
+  ExplicitRetirementHandle deferRetirement(String tag,
+                                           EntityMessage originalMessageToDefer,
+                                           EntityMessage futureMessage);
+
+  /**
    * Asynchronously send a message to the entity instance which looked up the service instance but also blocks the final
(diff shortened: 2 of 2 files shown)


standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
```diff
@@ -39,6 +39,18 @@
   void messageSelf(EntityMessage message) throws MessageCodecException;
 
   /**
+   * Defer a message based on another message, which will be sent in the future.
+   *
+   * @param tag String reason tag, mainly for debugging.
```
ChrisGreenaway: What is it for other than debugging?

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
```diff
@@ -39,6 +39,18 @@
   void messageSelf(EntityMessage message) throws MessageCodecException;
 
   /**
+   * Defer a message based on another message, which will be sent in the future.
```
ChrisGreenaway: I don't really understand *why* someone would want to do this. Can you give an example please?

standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *  The Initial Developer of the Covered Software is
+ *  Terracotta, Inc., a Software AG company
+ *
+ */
+package org.terracotta.entity;
+
+import java.util.concurrent.TimeUnit;
+
+public interface ExplicitRetirementHandle {
```
ChrisGreenaway: Perhaps this could extend AutoCloseable (or Closeable if we have to be Java 6)

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
cschanck: Generally, the requirement from TCStore/Ehcache was to be able to defer retirement to some point in the future, under entity control. We chose to implement this as a decomposition of the existing deferment, which was "here are these 2 messages, one depends on this other one, and send the other one now". This basically gives you a way to setup the deferment dependency without having to send the later message yet. 

Shorter answer, this was a feature we were explicitly asked for.

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
cschanck: Well, since this feature can leave to resource starvation, one can conceive of cases where you'd want to dump the list of outstanding explicit deferments, to see where you screwed up. This was a nice and clean way to denote "why" you did this. 

You can also foresee, if this becomes a problem, having the server monitor the outstanding explicit handles, and dump them to the log when they are a certain age, or even implementing a finalizer-esque  feature to release handles that have been GC'd. Or at least complain loudly about them. A small amount of context will go a long way.

standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
cschanck: We do have to java 6, sadly. Not sure if the meaning "release and finally send the message" matches up all that well with close(). But it could.

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
ChrisGreenaway: This hasn't cleared it up for me at all.

I don't understand the use case from EhCache or TCStore.

I don't understand what it even means to defer retirement. I thought messages are "retired" once they are delivered to all servers. How do you defer that?

Why is there a release() method if the message is always retired when the second one is sent?

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
ChrisGreenaway: So can we change "mainly for debugging" to "for debugging", or "for debugging and reporting"?

standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
ChrisGreenaway: OK - Java 6 - so Closeable? This has the advantage of being AutoCloseable on Java 7/8.

PR #218, opened 2017-05-12
Close #217 : add getServices

### LINKED ISSUE #217: Provide the ability to return multiple Service instances from ServiceRegistry
There may be instances where an entity would like to use multiple services implementing the same interface.  Provide the ability to return a collection of services based on a configuration

anthonydahanne:
Naive suggestion to have getServices and let MnM have access to several ManagementProviders

comments welcome !


myronkscott:
This change requires  a significant change in how services are registered and bootstrapped as well as how getService works. I am currently working on this issue. Please allow me to complete my work.

myronkscott:
using https://github.com/Terracotta-OSS/terracotta-apis/pull/221 instead

PR #219, opened 2017-05-12
Change ServerEntity invoke API to throw EntityUserException

ramsai1729:
This change is needed for Terracotta-OSS/terracotta-core#509


entity-server-api/pom.xml
```diff
@@ -40,5 +40,10 @@
       <artifactId>entity-common-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
myronkscott: this new dependency worries me a bit.  I wonder if a new exception type should be defined on the server-side that gets translated to EntityUserException on the client side.

cschanck: It does seem like it might be better to create an entirely new exception, rather than stretch EntityUserException to fit.

myronkscott: I know it was my suggestion to use EntityUserException but I think we need a new Exception type in entity-common-api next to MessageCodeException.  What do you think?

PR #228, opened 2017-05-26
[DONT MERGE] Proposal for HA/Message Tracking changes

cschanck:
We have mostly decided on, for each invoke (active or passive) passing in the ClientDescriptor (active already did this) and the current and eldest message id. The current is used to track the invoke within the Client, the eldest is used for garbage collection.

Additionally, a new method is added to CommonServerEntity, which notifies at points where the client
view of the data is in synch with the servers. In particular, this would be used when a client disconnects, or when client reconnect happens. At both of these points all historical information related to the client can be thrown away.

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
@@ -52,11 +52,17 @@ public interface ActiveServerEntity<M extends EntityMessage, R extends EntityRes
    *  called concurrently with other invokes.</p>
    *
    * @param clientDescriptor source instance from which the invocation originates.
-   * @param message The message from the client
+   * @param currentOrderedId current id for this client that this invoke is part of
+   * @param eldestOrderedId earliest active id for this client; id's older than this
+   * can be thrown away.
+   * @param message The message from a client
    * @return possible return value
    */
-  R invoke(ClientDescriptor clientDescriptor, M message) throws EntityUserException;
-  
+  R invokeActive(ClientDescriptor clientDescriptor,
+           long currentOrderedId,
+           long eldestOrderedId,
+           M message) throws EntityUserException;
+
   /**
    * <p>Called when an entity was loaded from some persistent state and the entity is expected to already be known to the
    *  server.</p>
--- a/entity-server-api/src/main/java/org/terracotta/entity/CommonServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/CommonServerEntity.java
@@ -46,4 +46,5 @@ public interface CommonServerEntity<M extends EntityMessage, R extends EntityRes
    *  respect to all other messages enqueued for the entity.</p>
    */
   void destroy();
+
 }
--- a/entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
@@ -36,9 +36,16 @@ public interface PassiveServerEntity<M extends EntityMessage, R extends EntityRe
    *  under a given key, the messages it creates and sends for synchronization will be executed in that key, on the passive,
    *  via this invoke() entry-point.</p>
    *
+   * @param clientDescriptor source instance from which the invocation originates.
+   * @param currentOrderedId current id for this client that this invoke is part of
+   * @param eldestOrderedId earliest active id for this client; id's older than this
+   * can be thrown away.
    * @param message The message from a client or upstream active server
    */
-  void invoke(M message) throws EntityUserException;
+  void invokePassive(ClientDescriptor clientDescriptor,
+              long currentOrderedId,
+              long eldestOrderedId,
+              M message) throws EntityUserException;
 
   /**
    * <p>Called on {@link ConcurrencyStrategy#MANAGEMENT_KEY} to notify the receiver that it is about to start receiving
@@ -69,4 +76,12 @@ public interface PassiveServerEntity<M extends EntityMessage, R extends EntityRe
    *  where this call is executed.</p>
    */
   void endSyncConcurrencyKey(int concurrencyKey);
+  
+  /**
+   * Notify an entity that a particular client disconnected from the active.
+   *
+   * @param client client
+   */
+  void notifyClientDisconnectedFromActive(ClientDescriptor client);
+
 }


entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
```diff
@@ -36,9 +36,15 @@
    *  under a given key, the messages it creates and sends for synchronization will be executed in that key, on the passive,
    *  via this invoke() entry-point.</p>
    *
+   * @param clientDescriptor source instance from which the invocation originates.
+   * @param currentOrderedId current id this invoke is part of
+   * @param eldestOrderedId earliest active id for this client
    * @param message The message from a client or upstream active server
    */
-  void invoke(M message) throws EntityUserException;
+  void invoke(ClientDescriptor clientDescriptor,
```
mathieucarbou: Is it possible to have an example use case to see how to properly use the new 3 parameters ? I don't really understand how I would use this ClientDescriptor from a passive entity and I am wondering how to adapt the voltron-proxy the best way to support that.

1. what I can do, what I cannot do ? 
2. Is it the same (i.e. equals / hashcode) as the one on the active ? 
3. Do the passive entity will also be aware of the join and leave of this client descriptor ?
4. Do I have the ability to use it to store some things related to this descriptor, and have a way to clean that (meaning knowing that the descriptor is gone) ?
5. On a failover, for example, does the ClientDescriptor will remain the same on the new active as it was observed on the passive that became active ?

albinsuresh: Would there be any relation between a deferred message and its corresponding scheduled message sent using `IEntityMessenger` in terms of their `clientDescriptor`, `currentOrderedId` and `eldestOrderedId`?

entity-server-api/src/main/java/org/terracotta/entity/CommonServerEntity.java
```diff
@@ -46,4 +46,12 @@
...
    */
   void destroy();
+
+  /**
+   * Notify an entity that a particular client has a shared understanding of the state
+   * of the system, such as after the client disappears, or after reconnect has finished.
+   *
+   * @param client client
+   */
+  void notifyClientIsSynched(ClientDescriptor client);
```
cschanck: The action will be the same -- it will be called on disconnect (client isn't there anymore) and sometime after reconnect (when the client can be asserted to have caught up). In either case, you can throw away any tracking history, because we are asserting the client has the same picture as the server.

entity-server-api/src/main/java/org/terracotta/entity/PassiveServerEntity.java
cschanck: The additional information allows you to associate messages (invokes) with a specific client/transaction. For an entity, this can be used to handle the reconnect case. On reconnect (failover from active to passive), the client will resend all it's outstanding messages. It is possible for the client to resend a message that the newly promoted active has already applied. This information will allow for entities to track ignore already applied messages, if that is important. I believe Albin & Abhilash are building the common tracking functionality on top of this. 

Also, this is only of concern if your entity has non-idempotent operations. If you are fine with reapplying some messages twice on failover, then you can completely ignore this.

cschanck: Client Descriptor will be the same, current and eldest are not necessarily related.

mathieucarbou: thanks @cschanck for the explanations!

myronkscott: @albinsuresh. If I understand the question correctly, it's up to the entity to create and maintain the association with the message it creates for ientitymessenger.

albinsuresh: @myronkscott @cschanck yes that's what I wanted to know. Currently for the Ehcache entity we are creating the scheduled message with the original message's id and client id. Just wanted to confirm if that still needs to be done or I can directly use the ids coming in the invoke. And I got the answer. The client descriptor will be preserved and not the message ids.

entity-server-api/src/main/java/org/terracotta/entity/CommonServerEntity.java
albinsuresh: Okay now I get it. In the reconnect case this method is NOT invoked immediately after the handleReconnect call on the active BUT only after the client is done resending all the messages in its resend queue as well. So that's when the reconnect is really "finished". In that case, yes that information is probably relevant.

albinsuresh:
Since we have `notifyClientIsSynched` which signifies that a client has disconnected(or reconnected) it would have been nice if you could provide a `notifyClientConnected` as well so that the life cycle of a tracker is cleaner. We could create a tracker on `notifyClientConnected` and remove the tracker on `notifyClientIsSynched`.

Of course we can lazily create the tracker on the first invoke using a `ClientDescriptor`. `notifyClientConnected` is just a good to have addition if adding that is not too much of a work.

entity-server-api/src/main/java/org/terracotta/entity/CommonServerEntity.java
cschanck: Yeah. It wouldbe incumbent on the server to ensure it is called in a method that guarantees it is useful/reliable. Disconnect for sure is easy, because once the client is gone, we don't need the history. reconnect may be tougher, we will see. There may also be other cases where it could be done.

cschanck:
@albinsuresh Not sure about that, we'd have to guarantee that the call to clientConnected(ClientDescriptor) completed before any messages arrived for that ClientDescriptor, otherwise you'd have to do the lazy init anyway. I am not sure how easy that would be to do.

cschanck:
Closed by accident, still here.

cschanck:
Changed the isSynched() call to merely state the client is now invalid and will not be reused. Only the disconnect case is useful and can be done deterministically, so...

ChrisGreenaway:
@cschanck said: "we'd have to guarantee that the call to clientConnected(ClientDescriptor) completed before any messages arrived for that ClientDescriptor"

Presumably you could do this by invoking clientConnected on the management key.

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -52,11 +52,16 @@
...
    * @param clientDescriptor source instance from which the invocation originates.
+   * @param currentOrderedId current id this invoke is part of
+   * @param eldestOrderedId earliest active id for this client
    * @param message The message from the client
    * @return possible return value
    */
-  R invoke(ClientDescriptor clientDescriptor, M message) throws EntityUserException;
-  
+  R invoke(ClientDescriptor clientDescriptor,
+           long currentOrderedId,
```
ChrisGreenaway: I wonder whether we could / should combine the extra parameters into a single object. This would reduce the noise in the method signature and also allow more flexibility for the future.

One wrinkle is that the passive wants that object to have the triple (clientDescriptor, currentOrderedId, eldestOrderedId), whereas the active already has the clientDescriptor.

PR #229, opened 2017-06-09
Refactor State Dump API

ramsai1729:
* Rename StateDumper to StateDumpCollector
   * StateDumpCollector: Provide a special key to add state in JSON format


entity-common-api/src/main/java/org/terracotta/entity/StateDumpCollector.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  void addState(String key, String value);
+
+  /**
+   * Adds given key and json string value mapping to this {@link StateDumpCollector}
+   *
+   * @param key            key name with the value to be associated
+   * @param jsonString     the value which should be a valid json
+   */
+  void addJsonState(String key, String jsonString);
```
ChrisGreenaway: Not sure why we've got JSON value as input. I thought the idea was to abstract away what the output would be. If we were happy with JSON input we could just have the jsonString contain all the state for a StateDumpable, but I presume that's not the intention.

I can, however, see the value in supporting more than key -> value, like the array stuff that Chris D was talking about.

mathieucarbou: I disagree with this `String` type also because it cannot enforce any JSON string (use could still put XML or anything else), and it also does not take in consideration that some json string could be invalid. For example, what if the json string is null, or a json representation of a primitive, or an json array, or a json object: would all these work ?

If we are to use JSON, why can't we then *really* use JSON and use the JSON-P javax api for that ?

ramsai1729: @mathieucarbou we are good as long as given json string is valid (as the API expects). JSON-P is not part of jdk and we don't need to add some external dependency for entity api if we go string way. AFAIK we just need to log state information or dump in some file, no need to interpret (other than for pretty printing). Let me know we need to support some other use cases.

ramsai1729:
Updated PR with changes discussed in sync up today.

entity-common-api/src/main/java/org/terracotta/entity/StateDumpCollector.java
mathieucarbou: @ramsai1729 : so it means we can put anything in the string then if it is not interpreted ? like YAML for example ?

PR #238, opened 2017-06-15
Issue237: add Serializable to ClientDescriptor

cschanck:
Adds serializable, and isValid(), to ClientDescriptor. Active server services trying to use the message tracking may need to pass ClientDescriptors to passives, so Serializable seems necessary. 

isValid() is merely to represent the fact there exists a NULL implementation that means nothing, seems useful in order to ignore client descriptors that are ersatz.

### DIFF
--- a/entity-server-api/src/main/java/org/terracotta/entity/ClientDescriptor.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/ClientDescriptor.java
@@ -18,11 +18,15 @@
  */
 package org.terracotta.entity;
 
+import java.io.Serializable;
+
 /**
  * An opaque token representing a specific entity instance on a specific client node.
- * This is used by server-side code to specifically communicate with or track connection status of a specific client-side
- * object instance.
+ * This is used by server-side code to specifically communicate with or track connection
+ * status of a specific client-side object instance. This is used both in actives
+ * and passives in certain cases.
  * Note that implementations are expected to provide a proper equals()/hashCode().
  */
-public interface ClientDescriptor {
+public interface ClientDescriptor extends Serializable {
+
 }
--- a/entity-server-api/src/main/java/org/terracotta/entity/InvokeContext.java
+++ b/entity-server-api/src/main/java/org/terracotta/entity/InvokeContext.java
@@ -22,4 +22,11 @@ public interface InvokeContext {
    * @return txid, or -1 if unknown.
    */
   long getOldestTransactionId();
+
+  /**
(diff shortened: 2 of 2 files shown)


entity-server-api/src/main/java/org/terracotta/entity/ClientDescriptor.java
```diff
@@ -18,11 +18,20 @@
...
  * Note that implementations are expected to provide a proper equals()/hashCode().
  */
-public interface ClientDescriptor {
+public interface ClientDescriptor extends Serializable {
+  /**
+   * Does this ClientDescriptor represent an actual client endpoint,
+   * or is it merely a marker.
+   * @return
+   */
+  boolean isValid();
```
ljacomet: Is this really something a user of the server API needs to care about?
If yes, what is supposed to happen when an entity finds itself handled an invalid `ClientDescriptor`?

cschanck:
Yeah, I think so. Non valid descriptors (NULL static in the impl) are
passed instead of null when things are unrelated to a real client. We do
not pass null, but an invalid descriptor means you can't do any meaningful
correlation with it. Passive sync, lots of config stuff, etc, all push
invalid descriptors.

For purposes of tracking, for example, these need to be ignored.

And no, changing it optional is not a real answer, and brother is actually
passing null. 😀😀

entity-server-api/src/main/java/org/terracotta/entity/ClientDescriptor.java
cschanck: I would, however enjoy suggestions for better naming.

henri-tremblay: isValidEndpoint, isARealEndpoint, isReal, isMarker, isExistingEndpoint, isRepresentingRealEndpoint ?

myronkscott: I wonder if we should tag the InvokeContext with isServerGenerated() instead.

cschanck:
Well, it is ClientDesxriptor.isValid, so adding CkientDescriptor seems a
bit excessive.

entity-server-api/src/main/java/org/terracotta/entity/ClientDescriptor.java
cschanck: @myronkscott I kinda hate that, as it introduces a bug path, where a mistake is made and the two are wrong.

PR #245, opened 2017-07-21
Add disconnectFinished() and ClientDescriptor ser/desr support in InvokeContext

cschanck:


entity-server-api/src/main/java/org/terracotta/entity/InvokeContext.java
```diff
@@ -29,4 +29,18 @@
    * @return true if the client information represents a real client.
    */
   boolean isValidClientInformation();
+
+  /**
+   * Create a tight byte representation of a client descriptor
+   * @param clientDescriptor
+   * @return
+   */
+  byte[] getClientDescriptorAsBytes(ClientDescriptor clientDescriptor);
```
ChrisGreenaway: Or perhaps allow passing a ByteBuffer, to avoid copying.

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -104,6 +104,12 @@ default R invoke(ClientDescriptor descriptor, M message) throws EntityUserExcept
   void handleReconnect(ClientDescriptor clientDescriptor, byte[] extendedReconnectData) throws ReconnectRejectedException;
 
   /**
+   * <p>Called when a reconnect event has finished. Either all clients have reconnected, or
+   * the reconnect window has timed out.</p>
+   */
+  void reconnectFinished();
```
ChrisGreenaway: Can you add to the Javadoc something about how this is called in relation to invokeActive()?

In particular, I would expect that invokeActive() won't be called until after reconnectFinished() is called.

myronkscott:
went a different path

PR #247, opened 2017-07-24
Rework InvokeContext to add new ClientSourceId

cschanck:


entity-server-api/src/main/java/org/terracotta/entity/InvokeContext.java
```diff
@@ -1,12 +1,29 @@
...
 public interface InvokeContext {
   /**
-   * source instance from which the invocation originates.
-   *
-   * @return descriptor
+   * Source instance from which the invocation originates.
+   * @return client source
    */
-  ClientDescriptor getClientDescriptor();
+  ClientSourceId getClientSource();
```
ChrisGreenaway: This is a massive breaking change.

PR #259, opened 2017-08-14
Fix #254: Make EndpointDelegate generic

### LINKED ISSUE #254: EndpointDelegate should have generic type bounds
If ```EndpointDelegate``` became ```EndpointDelegate<R extends EntityResponse>``` then ```handleMessage(EntityResponse entityResponse)``` could be ```handleMessage(R r)```.

ramsai1729:


entity-client-api/src/main/java/org/terracotta/entity/EntityClientEndpoint.java
```diff
@@ -44,7 +44,7 @@
    * 
    * @param delegate The delegate to use for events originating within the receiver.
    */
-  void setDelegate(EndpointDelegate delegate);
```
myronkscott: This set method looks like an implementation detail and probably should be removed form the API

myronkscott: Nevermind.

Issue #263, opened 2017-08-24
`RuntimeEntityException` subclasses should require entity name and type info.


chrisdennis:
In order to avoid accidental omission of entity name and type information from exception messages we shouldn't have constructors for `RuntimeEntityException` subclasses that don't take entity and type information.

myronkscott:
Addressed by Terracotta-OSS/terracotta-core#726.  All exceptions have the type and name information added to them when thrown locally on the client

chrisdennis:
The improvements are good... but the exception that led me to this was `ConnectionClosedException`.  Re-opening to see if that one can be fixed.

myronkscott:
Unfortunately, in order to fix that one, the exception is re-wrrapped into LocalConnectionClosedException.  Are you sniffing the exception stack or just checking the type and name of the top level exception?

chrisdennis:
My bad... should have looked closer at the changeset.

PR #265, opened 2017-08-24
Issue #263 : Enforce non-nullness of entity information in entity exceptions

chrisdennis:


ljacomet:
Might be worth fixing that on both `release/1.3` and `master`

chrisdennis:
@myronkscott @cschanck thoughts on this?  If you are okay with the principle here I'll convert it in to a proper daggy fix.

myronkscott:
I don't think this is a good idea to port back to 1.3.  It is a HUGE implementation change and might involve message changes.

chrisdennis:
Is it good for 1.4 though?

myronkscott:
I hope so but I need some time to investigate.  The first place I spot checked looked like a complicated change which inspired the last comment.

chrisdennis:
@myronkscott any good news to tell me?

myronkscott:
@chrisdennis this will be handled via implementation in terracotta-core and available in the next release.

myronkscott:
addressed by Terracotta-OSS/terracotta-core#726

PR #271, opened 2017-09-27
Convert to java8

myronkscott:


packaging-support/src/main/java/com/tc/classloader/PermanentEntity.java
```diff
@@ -30,8 +31,10 @@
  */
 @Target( ElementType.TYPE )
 @Retention( RetentionPolicy.RUNTIME )
+@Repeatable( Permanents.class )
 public @interface PermanentEntity {
   String type();
-  String[] names();
+  String[] names() default {};
```
ramsai1729: what is the use of names() now?

packaging-support/src/main/java/com/tc/classloader/OverrideService.java
```diff
@@ -19,20 +19,22 @@
...
 
 /**
  * Tagging a Service type with an override causes the classloader to ignore the 
- * overridden types for the type provided.  Each type is the fully qualified classname
- * of the type to be overridden.  The overridden types the and type of the interface 
- * tagged should be the same from the perspective of the classloader.
+ * overridden types for the type provided.  The value is a fully qualified class name 
+ * which is to be overridden.
  * 
  * TODO: consider using the @Repeatable once java 6 support is dropped
```
ramsai1729: this comment can be removed now.

packaging-support/src/main/java/com/tc/classloader/PermanentEntity.java
myronkscott: Backward compatible.  You can now do either 

@PermanentEntity(type=type,names=[one,two])
public class EntityServerService

or

@PermanentEntity(type=type,names=one)
@PermanentEntity(type=type,names=two)
public class EntityServerService

packaging-support/src/main/java/com/tc/classloader/OverrideService.java
myronkscott: updated

PR #276, opened 2017-10-23
Add response callbacks to IEntityMessenger API

myronkscott:
@ljacomet @AbfrmBlr @chrisdennis - Please review if you find useful or not.


chrisdennis:
@AbfrmBlr Could you hack on this some to see if it makes sense?

PR #277, opened 2017-10-25
Add consumer to retirement handle

myronkscott:


myronkscott:
@AbfrmBlr please review for usability

standard-cluster-services/src/main/java/org/terracotta/entity/ExplicitRetirementHandle.java
```diff
@@ -18,11 +18,13 @@
...
+import java.util.function.Consumer;
+import org.terracotta.entity.IEntityMessenger.MessageResponse;
 
-public interface ExplicitRetirementHandle {
+public interface ExplicitRetirementHandle<R extends EntityResponse> {
   String getTag();
 
   void release() throws MessageCodecException;
 
+  void release(Consumer<MessageResponse<R>> consumer) throws MessageCodecException;
```
AbfrmBlr: Below is the use case for Long running operations. I am trying to create a typical flow of requests/responses which are involved while server executes a Long Running Operation(LRO).

Client | Active
-------- | -------- 
Sends **M1** | Receives **M1**
Blocked    | Forks a **LRO** in separate thread, & associates an **ID** with that operation
Blocked    | Defers Retirement of **M1** with **M1'** & gets handle **H**
Blocked    | Sends Response **R1** with **ID**  
Blocked    | After sometime, when **LRO** completes & releases handle **H**
Receives **R1** with **ID** | -----------
Sends **M2** with **ID** | Looks up **LRO** result with **ID**, & sends reponse **R2**
Receives **R2** which is the actual result | -----------

In above flow, it would be really helpful, if we can somehow eliminate the need of two server round trips for acquiring the actual result of a LRO.

One way to achieve that is by sending **R2** to the actual client as the response to **M1** when we are releasing the handle itself.

chrisdennis: This makes sense, but it's a request for deferred response.  It's valid and I'd like to see it achieved, but it's not within the scope of this issue.  This is supposed to make the code that ensures the LRO wait for retirement (aka passive receive) of any messages sent by the LRO during it's operation before it releases the handle `H`.

myronkscott:
Handled by other means.

PR #278, opened 2017-10-25
Reconnect sub-lifecycle

myronkscott:


myronkscott:
@ljacomet @AbfrmBlr @chrisdennis - please review

entity-server-api/src/main/java/org/terracotta/entity/ActiveServerEntity.java
```diff
@@ -125,4 +111,37 @@ default void prepareKeyForSynchronizeOnPassive(PassiveSynchronizationChannel<M>
    * @param concurrencyKey The key of the data to be synchronized
    */
   void synchronizeKeyToPassive(PassiveSynchronizationChannel<M> syncChannel, int concurrencyKey);
+  /**
```
cschanck: Oh that's very nice indeed.

PR #281, opened 2017-11-02
Small fixes

myronkscott:
Better EntityRuntimeException handling.
Clarification in javadoc


connection-api/src/main/java/org/terracotta/exception/RuntimeEntityException.java
```diff
@@ -51,7 +51,7 @@ public String getDescription() {
    * @param cause The underlying cause of the exception, null if nothing appropriate
    */
   protected RuntimeEntityException(String className, String entityName, String description, Throwable cause) {
-    super("Entity: " + className + ":" + entityName + " " + description, cause);
+    super(className != null ? "Entity: " + className + ":" + entityName + " " + description : description, cause);
```
cschanck: can className!=null but entityName==null? what about description?

myronkscott:
obsolete

PR #283, opened 2017-11-20
Validate connection URIs

ChrisGreenaway:


ChrisGreenaway:
Any extra information that a ConnectionService needs (e.g. a username / password) would be passed via the Properties object.

Issue #285, opened 2018-01-31
Add maxSyncConcurrency() API to EntityServerApi


rkavanap:
This can default to return 1 and will allow active entities to provide hints on how much concurrent threads can be used to synchronize and provides an indication of the  amount of resource usage used by the entity to synchronize a passive entity.

PR #286, opened 2018-01-31
#285 : API to specify maxSyncConcurrency on active server entity

rkavanap:
@myronkscott .. Take a look and see if it can be merged.

Will raise another PR in core..if we agree this is worthwhile. I am ready with the code.


rkavanap:
Feature not approved...

PR #290, opened 2018-02-17
Issue#284: Added parallel APIs to ConnectionFactory.connect and ConnectionService.connect

saurabhagas:


connection-api/src/main/java/org/terracotta/connection/ConnectionFactory.java
```diff
@@ -56,4 +71,17 @@ private static void validateURI(URI uri) throws ConnectionException {
       throw new ConnectionException(e);
     }
   }
+
+  private static Connection findConnectionService(List<String> servers, Properties properties)
+      throws ConnectionException {
+    String scheme = properties.getProperty(ConnectionPropertyNames.CONNECTION_SCHEME, "terracotta");
```
mathieucarbou: I don't understand how it would work for management, which needs to use different schemes to connect to a cluster. We are using: `terracotta://`, `diagnostic://` and `stripe://` (see downstream projects).

mathieucarbou: @anthonydahanne @GaryWKeim : FYI

mathieucarbou: UPDATE: Oh I see! We have to pass the scheme as a property.... @anthonydahanne : this will be an important change for us to keep in mind!

saurabhagas:
Requesting comments..

connection-api/src/main/java/org/terracotta/connection/util/HostPort.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.terracotta.connection.util.HostPortValidator.isValidHost;
+import static org.terracotta.connection.util.HostPortValidator.isValidIPv4;
+import static org.terracotta.connection.util.HostPortValidator.isValidIPv6;
+import static org.terracotta.connection.util.HostPortValidator.isValidPort;
+
+/**
+ * A utility class to store the server host port information.
+ */
+public class HostPort {
+  private final String host;
```
myronkscott: This seems like it should be part of implementation if the API calls for string

connection-api/src/main/java/org/terracotta/connection/util/HostPortValidator.java
```diff
@@ -0,0 +1,75 @@
+package org.terracotta.connection.util;
+
+/**
+ * A class which provides utility methods to test the validity of IP address, host name and port.
+ */
+
+public class HostPortValidator {
+  private final static String VALID_HOSTNAME = "^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$";
```
myronkscott: This should be part of the implementation library

connection-api/src/main/java/org/terracotta/connection/ConnectionService.java
```diff
@@ -52,4 +65,15 @@
...
+  /**
+   * Establish a connection to the given list of servers using the specified properties. handlesScheme() must be called
+   * with the scheme prior to connect(). Calling connect() on an unverified scheme can lead to unspecified behavior.
+   *
+   * @param servers list of servers to connect to
+   * @param properties user specified implementation specific properties
+   * @return established connection
+   * @throws ConnectionException on connection failure
+   */
+  Connection connect(List<String> servers, Properties properties) throws ConnectionException;
```
myronkscott: Can this be `connect(String scheme, Iterable<InetSocketAddress> listOfServers, Properties cps)` or `connect(SomeEnumType scheme, Iterable<InetSocketAddress> listOfServers, Properties cps)`?

connection-api/src/main/java/org/terracotta/connection/ConnectionPropertyNames.java
```diff
@@ -35,4 +35,6 @@
   String CONNECTION_DISABLE_RECONNECT = "connection.noreconnect";
   /*  request a specific connection type */
   String CONNECTION_TYPE = "connection.type";
+  /*  request a specific connection scheme */
+  String CONNECTION_SCHEME = "connection.scheme";
```
myronkscott: Probably makes more sense as an argument in connect.  If this stays as a property, there is overlap here with CONNECTION_TYPE.  Either make this an argument in connect or merge your additional definitions into CONNECTION_TYPE

connection-api/src/main/java/org/terracotta/connection/ConnectionService.java
saurabhagas: Should `ConnectionFactory.connect` have a similar signature too then?
1. If yes, the upper layers will have to directly pass `Iterable<InetSocketAddress> listOfServers`. Seems somewhat inconvenient, but would save us the validation effort.
2. If no, should it accept a `List<String> servers` and parse the `String`s into `InetSocketAddress`es?

connection-api/src/main/java/org/terracotta/connection/ConnectionService.java
```diff
@@ -42,6 +47,14 @@
    */
   boolean handlesURI(URI uri);
```
saurabhagas: Can this go away in favor of `handlesScheme`?

GaryWKeim:
I have a concern that this will be a backward-incompatible change. Are we planning on leaving the user-interface unchanged by parsing the user-provide connection string into the List<String || InetAddress> that can be passed to ConnectionService?

PR #294, opened 2018-07-04
Migrated the tests (except integration tests which depend on Galvan, …

bbyreddy:
…Angela or Passthrough frameworks) to Junit 5.

### DIFF
--- a/connection-api/src/test/java/org/terracotta/exception/RuntimeEntityExceptionTest.java
+++ b/connection-api/src/test/java/org/terracotta/exception/RuntimeEntityExceptionTest.java
@@ -5,12 +5,12 @@
  */
 package org.terracotta.exception;
 
-import org.junit.After;
-import org.junit.AfterClass;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
 /**
  *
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -53,6 +53,12 @@
         <artifactId>guava</artifactId>
         <version>18.0</version>
       </dependency>
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-engine</artifactId>
+        <version>5.2.0</version>
+        <scope>test</scope>
```
lorban: There shouldn't be a scope in `dependencyManagement` dependencies.

connection-api/src/test/java/org/terracotta/exception/RuntimeEntityExceptionTest.java
```diff
@@ -57,11 +57,11 @@ public void testEntityInfoProvided() {
...
-    Assert.assertNotEquals(description, impl.getMessage());
-    Assert.assertEquals(clz, impl.getClassName());
-    Assert.assertEquals(name, impl.getEntityName());
-    Assert.assertTrue(impl.getMessage().contains(clz));
-    Assert.assertTrue(impl.getMessage().contains(name));
+    Assertions.assertNotEquals(description, impl.getMessage());
+    Assertions.assertEquals(clz, impl.getClassName());
+    Assertions.assertEquals(name, impl.getEntityName());
+    Assertions.assertTrue(impl.getMessage().contains(clz));
+    Assertions.assertTrue(impl.getMessage().contains(name));
```
lorban: Static imports would look nicer.

pom.xml
bbyreddy: Removed the scope from the dependencyManagement in this and other pom files as well.

connection-api/src/test/java/org/terracotta/exception/RuntimeEntityExceptionTest.java
bbyreddy: Fixed

bbyreddy: Dependency on multiple versions of Junit will exists temporarily, only till we migrate the Galvan tests as well.

PR #297, opened 2018-08-02
discover properties of the client source if available

myronkscott:


entity-server-api/src/main/java/org/terracotta/entity/ActiveInvokeContext.java
```diff
@@ -34,5 +36,7 @@
    * @return a channel to send messages to the originator of the message associated with this context
    */
   ActiveInvokeChannel<R> openInvokeChannel();
+  
+  Properties getClientSourceProperties();
```
ChrisGreenaway: Wondering whether we could use a better interface than Properties (e.g. Map<String, Object>)

myronkscott: This is informational.  I don't think we should expose any more than strings much like header information in http.

ChrisGreenaway: So if it's just a string to string mapping then Map<String, String>, unlike Properties which is Map<Object, Object>.

But, I thought one of the properties would contain a BufferManager as a value. Is that not right?

PR #304, opened 2018-11-06
Api additions

myronkscott:
multiple additions to the API for new features.  No breaking changes to the current API.

### DIFF
--- a/connection-api/src/main/java/org/terracotta/exception/ConnectionClosedException.java
+++ b/connection-api/src/main/java/org/terracotta/exception/ConnectionClosedException.java
@@ -28,14 +28,16 @@ package org.terracotta.exception;
  */
 public class ConnectionClosedException extends RuntimeEntityException {
   private static final long serialVersionUID = 1L;
+  private final boolean wasSent;
 
   /**
    * Creates a new instance with the given description.
    * 
    * @param description The description of the exception.
    */
-  public ConnectionClosedException(String description) {
+  public ConnectionClosedException(boolean wasSent, String description) {
     super(null, null, description, null);
+    this.wasSent = wasSent;
   }
 
   /**
@@ -44,7 +46,18 @@ public class ConnectionClosedException extends RuntimeEntityException {
    * @param description The description of the exception.
    * @param cause The underlying cause of the exception.
    */
-  public ConnectionClosedException(String description, Throwable cause) {
+  public ConnectionClosedException(boolean wasSent, String description, Throwable cause) {
     super(null, null, description, cause);
+    this.wasSent = wasSent;
+  }
+  
+  /**
+   * Designates whether the message associated with this exception was sent to the cluster.
+   * 
+   * @return true of the message was sent to the server
+   */
+  
+  public boolean messageWasNotSent() {
+    return !this.wasSent;
   }
 }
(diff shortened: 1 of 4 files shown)


connection-api/src/main/java/org/terracotta/exception/ConnectionClosedException.java
```diff
@@ -44,7 +46,18 @@ public ConnectionClosedException(String description) {
...
+    this.wasSent = wasSent;
+  }
+  
+  /**
+   * Designates whether the message associated with this exception was sent to the cluster.
+   * 
+   * @return true of the message was sent to the server
+   */
+  
+  public boolean messageWasSent() {
```
chrisdennis: The sense of this seems wrong. Definite action should only be taken here when this method returns false.  Essentially the return values here are really: `false` and `maybe`. I'm wondering if an inverted method might lead to less mistakes?

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
```diff
@@ -31,6 +31,12 @@
  */
 @CommonComponent
 public interface IEntityMessenger<M extends EntityMessage, R extends EntityResponse> {
+  /**
+   * Sends a message to delete the entity.  Must be called from the 
+   * {@link org.terracotta.entity.ActiveServerEntity#disconnected(org.terracotta.entity.ClientDescriptor) disconnected}
```
chrisdennis: Presumably we can also call from other places too?  What we're really saying is - destroy will fail if clients are still connected... it's just much more likely that no clients will be connected when calling from disconnected (when the last client gets disconnected).

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -36,6 +36,13 @@
    * Stops the server
    */
   void stopPlatform();
+  
+  /**
+   * Server uptime.
+   * 
+   * @return the amount of milliseconds that have passed since the server was started
```
chrisdennis: Does started mean since the last state transition? What about zaps?

connection-api/src/main/java/org/terracotta/exception/ConnectionClosedException.java
myronkscott: makes sense. will fix

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
myronkscott: true

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
myronkscott: this is purely server start regardless of server state.

chrisdennis: So if a server gets zapped it's uptime restarts... it kinda leaks the way zaps happen but I guess that is relatively harmless.

standard-cluster-services/src/main/java/org/terracotta/entity/IEntityMessenger.java
myronkscott: It is actually guaranteed to fail if we call from anywhere else in the code due to the way we sequence a disconnect message.

PR #307, opened 2018-11-29
Add PlatformService extensions to implement shutdown primitives

ramsai1729:
@myronkscott These changes are same as the API changes shared by Ramesh earlier. Please review.

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
@@ -54,4 +54,42 @@ public interface PlatformService {
    * Provides the raw config file from server startup
    */
   InputStream getPlatformConfiguration();
+
+  /**
+   * Stops the server if it is in any of the passive states
+   * <p>
+   * Based on the {@code restartMode} argument, the server should either bounce or just halt.
+   * Currently, it is assumed that the halt may be abrupt and it may not clean up or empty any queues.
+   *
+   * @param restartMode mode in which the system needs to restart. STOP_ONLY means halt without restarting
+   *
+   * @throws PlatformStopException when platform stop fails for some reason
+   */
+  void stopPlatformIfPassive(RestartMode restartMode) throws PlatformStopException;
+
+  /**
+   * Stops the server if it is in any of the active states (including active suspended)
+   * <p>
+   * Based on the {@code restartMode} argument, the server should either bounce or just halt.
+   * Currently, it is assumed that the halt may be abrupt and it may not clean up or empty any queues.
+   *
+   * @param restartMode mode in which the system needs to restart. STOP_ONLY means halt without restarting
+   *
+   * @throws PlatformStopException when platform stop fails for some reason
+   */
+  void stopPlatformIfActive(RestartMode restartMode) throws PlatformStopException;
+
+  /**
+   * Stops the server.
+   * <p>
+   * Based on the {@code restartMode} argument, the server should either bounce or just halt.
+   * Currently, it is assumed that the halt may be abrupt and it may not clean up or empty any queues.
+   *
+   * @param restartMode mode in which the system needs to restart. STOP_ONLY means halt without restarting
+   */
+  void stopPlatform(RestartMode restartMode);
+
+  enum  RestartMode {
+    STOP_ONLY, STOP_AND_RESTART
+  }
 }
--- a/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformStopException.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformStopException.java
@@ -1,0 +1,13 @@
+package org.terracotta.monitoring;
+
+public class PlatformStopException extends Exception {
+  private static final long serialVersionUID = 7404140788018130043L;
+
+  public PlatformStopException(String message) {
+    super(message);
+  }
+
+  public PlatformStopException(String message, Throwable cause) {
+    super(message, cause);
+  }
+}


standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -54,4 +54,51 @@
...
+  /**
+   * Returns true if it is safe to stop.
+   * <p>
+   * {@code partialStop} is true means this is a rolling stop (for rolling upgrades etc). Otherwise, this is a full stripe shutdown
+   * <p>
+   * If partialStop is true and if this server is ACTIVE_COORDINATOR, then return true iff all other stripe members are in PASSIVE_STANDBY, false otherwise
+   * If partialStop is true and if this server is PASSIVE_STANDBY, then return true
+   * If partialStop is true and if this server is any other state, then return false
+   * If partialStop is false, return true always (for now)
+   *
```
myronkscott: Is this necessary?  It is not giving you good information since it is incredibly transient.  Anything in this area should be rolled into the operations above where the implementation can properly deal with it.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -54,4 +54,51 @@
...
+   */
+  boolean isSafeToStop(boolean partialStop);
+
+  enum  RestartMode {
+    STOP_ONLY, STOP_AND_RESTART_IN_SAFE_MODE, STOP_AND_RESTART
```
myronkscott: SAFE_MODE doesn't mean anything when the stripe is running and should be removed.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
rkavanap: partialStop true means it is only part of the stripe that is going down. In a planned shutdown case, it is important we fail the shutdown in case there is a possibility that system is currently not in a state (e.g mirror is not ready) to shutdown yet.

myronkscott: Understood but this check is not helpful as the state can change right after check, thus the next call taking place on information that is no longer true.  Any check like this should occur in implementation under one of the shutdown calls added above.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
rkavanap: this is for restarting/bouncing when an entire cluster is asked to stop and restart, 

@albinsuresh if you think this is not needed anymore (not sure about the outcome of all the restart discussions) even for AVAILABILITY failover mode, please let me know. We can remove this.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
rkavanap: ok..Yes that makes absolute sense. (currently since we halt a step after the check, the same hole may still exist..but having an integrated API makes it easier to close that window in the future).

@ramsai1729 can you integrate the stop above with this..

  boolean stopPlatform(RestartMode restartMode, boolean partialMode);

if partialMode is true and condition is not safe to shutdown, return false...
otherwise just halt..

rkavanap: @ramsai1729 Myself and @myronkscott  had a discussion..

So can you remove the isSafe... api  completely. 

There is no need of changes elsewhere including in stopPlatform...So ignore my comment above..

Throw an IllegalStateException if you cannot stop for the stop* calls. (so they can remain void)...

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
rkavanap: @ramsai1729, please remove the STOP_AND_RESTART_IN_SAFE_MODE from above.

ramsai1729: done

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
ramsai1729: done

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -54,4 +54,38 @@
...
+
+  /**
+   * Stops the server if it is in any of the passive states
+   * <p>
+   * Based on the {@code restartMode} argument, the server should either bounce or just halt (exit with correct exit codes).
+   * Currently, it is assumed that the halt may be abrupt and it may not clean up or empty any queues.
+   *
+   * @param restartMode mode in which the system needs to restart. STOP_ONLY means halt without restarting
+   */
+  void stopPlatformIfPassive(RestartMode restartMode);
```
myronkscott: should have a boolean return or throw a checked exception in the case that server is not in the proper state to shutdown

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -54,4 +54,38 @@
...
+  /**
+   * Stops the server if it is in any of the active states (including active suspended)
+   * <p>
+   * Based on the {@code restartMode} argument, the server should either bounce or just halt.
+   * Currently, it is assumed that the halt may be abrupt and it may not clean up or empty any queues.
+   *
+   * @param restartMode mode in which the system needs to restart. STOP_ONLY means halt without restarting
+   */
+  void stopPlatformIfActive(RestartMode restartMode);
+
```
myronkscott: should have a boolean return or throw a checked exception in the case that server is not in the proper state to shutdown

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
myronkscott: I think it needs to be a checked exception so that it is handled in the context which it is thrown.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
albinsuresh: We still have safe mode startup in availability mode.

ramsai1729: added STOP_AND_RESTART_IN_SAFE_MODE back

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
ramsai1729: added checked exception

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
ramsai1729: added checked exception

rkavanap:
@ramsai1729 LGTM.   Just double check with @myronkscott on the enum on restart mode. He seems to think that is no longer necessary, but Albin feels it is still needed in AVAILABILITY mode..

So there is a disconnect. I had also originally thought it is needed in AVAILABILITY mode.

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
myronkscott: I don't see any reasonable use-case scenario for this mode.  Please remove it and I will merge.

PR #312, opened 2019-06-25
client version in cluster state dumps.

GaryWKeim:


anthonydahanne:
@myronkscott can you please add your review ? thanks

myronkscott:
I have a concern that in rare cases, this change may crash the cluster.  If a stripe split-brains, then heals, the active will as the other active for information that includes this serialized object.  I know it is not a great solution, but maybe we should include version information as part of the UUID.

myronkscott:
I think the best way to handle this is to add a new sub-class of PlatformConnectedClient which includes the version.

anthonydahanne:
hello, @myronkscott do you want me or @GaryWKeim to add this new subclass ?

myronkscott:
> hello, @myronkscott do you want me or @GaryWKeim to add this new subclass ?

I think we need to discuss.  Even a subclass is bad as a legacy server will not have the sub-class and cause problems.

PR #315, opened 2020-04-09
Api evolution

myronkscott:
Move interfaces and references from implementation to API in an effort to formalize the interface of items required by those that are extending the platform.


connection-api/src/main/java/org/terracotta/connection/DiagnosticsConfig.java
```diff
@@ -0,0 +1,42 @@
...
+  public DiagnosticsConfig(Properties properties, Runnable close) {
+    this.properties = properties;
+    this.close = close;
+  }
+
+  public Properties getProperties() {
+    return properties;
+  }
+
+  public Runnable getClose() {
```
mathieucarbou, line 39: How this works ? What is responsible to call `getClose().run()` to close the connection created by `DiagnosticsFactory` ?

entity-server-api/src/main/java/org/terracotta/entity/ExecutionStrategy.java
```diff
@@ -36,6 +36,8 @@ public boolean runOnActive() {
       }
     }, BOTH {
 
+    }, ALL {
```
mathieucarbou, line 39: javadoc says: `on the active server, passive server or both`. So what `ALL` means ?

standard-cluster-services/src/main/java/org/terracotta/monitoring/PlatformService.java
```diff
@@ -90,6 +88,6 @@
   void stopPlatform(RestartMode restartMode);
 
   enum  RestartMode {
-    STOP_ONLY, STOP_AND_RESTART
+    STOP_ONLY, STOP_AND_RESTART, ZAP_AND_RESTART, ZAP_AND_STOP;
```
mathieucarbou, line 91: thanks 👍

standard-cluster-services/src/main/java/org/terracotta/server/Guardian.java
```diff
@@ -0,0 +1,51 @@
...
+package org.terracotta.server;
+
+import java.util.Properties;
+
+/**
+ *
+ */
+public interface Guardian {
+  public enum Op {
+    // these entity operations cannout be validated, they are only here
```
mathieucarbou, line 28: cannot

standard-cluster-services/src/main/java/org/terracotta/server/Server.java
```diff
@@ -0,0 +1,81 @@
...
+public interface Server {
+  int getServerCount();
+  
+  String[] processArguments();
+
+  void stop(StopAction...modes);
+
+  boolean stopIfPassive(StopAction...modes);
+
+  boolean stopIfActive(StopAction...modes);
```
mathieucarbou, line 36: `PlatformService` has the same methods / enum but in/with different classes.
Couldn't we have the same everywhere ?
Or Have `PlatformService` return a `Server` object and provide default implementations in the `PlatformService` interface if we really want backward compat' ?

standard-cluster-services/src/main/java/org/terracotta/server/Server.java
```diff
@@ -0,0 +1,81 @@
...
+
+  void dump();
+
+  String getResourceState();
+
+  String getClusterState();
+
+  String getConfiguration();
+
+  ClassLoader getServiceClassLoader(ClassLoader parent, Class<?>...serviceClasses);
```
mathieucarbou, line 72: Is that the new method I would need to use to create the ServiceClassLoader in the dynamic config `ConfigurationProvider` ?

standard-cluster-services/src/main/java/org/terracotta/server/ServerEnv.java
```diff
@@ -0,0 +1,49 @@
...
+public class ServerEnv {
+
+  private static Server defaultServer;
+  private static final ThreadLocal<Server> threadServer = new ThreadLocal<>();
+
+  public static void setDefaultServer(Server server) {
+    defaultServer = server;
+  }
+
+  public static Server getDefaultServer() {
```
mathieucarbou, line 33: Will I be able to eagerly access this method from our `ConfigurationProvider` ?

connection-api/src/main/java/org/terracotta/connection/DiagnosticsConfig.java
myronkscott: This has to be here for backwards compatibility and not created directly by the user.  The DiagnosticsFactory.connect is a convenience factory to create diagnostic connections.

standard-cluster-services/src/main/java/org/terracotta/server/ServerMBean.java
```diff
@@ -0,0 +1,39 @@
...
+  public static ObjectName createMBeanName(String name) throws MalformedObjectNameException {
+    return createMBeanName("org.terracotta", name);
+  }
+
+  public static ObjectName createMBeanName(String domain, String name) throws MalformedObjectNameException {
+    final StringBuilder objName = new StringBuilder(domain);
+    objName.append(':');
+    objName.append("name=").append(name);
+    return new ObjectName(objName.toString());
+  }
```
mathieucarbou, line 38: Couldn't we move these methods in `ServerJMX` and make them non static (eventually default) ? 
Do we really need to know about the impl here or its something that could be moved to the `ServerJMX` impl ?

entity-server-api/src/main/java/org/terracotta/entity/ExecutionStrategy.java
myronkscott: ALL and BOTH are synonyms

standard-cluster-services/src/main/java/org/terracotta/server/Server.java
myronkscott: I'd rather leave PatformService as is for backwards compatibility.  The Server global is a necessary evil and I don't want the service to reply on it.

standard-cluster-services/src/main/java/org/terracotta/server/Server.java
myronkscott: yes.  I have the changes for tc-platform queued once this is committed.

connection-api/src/main/java/org/terracotta/connection/DiagnosticsConfig.java
mathieucarbou: Ok so if I understand, no client should use `DiagnosticsFactory` because if so, it would have no way to properly close the created connection right ? These classes are just there for some reasons but we shouldn't use them ?

standard-cluster-services/src/main/java/org/terracotta/server/ServerEnv.java
myronkscott: yes

standard-cluster-services/src/main/java/org/terracotta/server/ServerMBean.java
myronkscott: This is convenience and backwards compatibility.  Implementors used the terracotta MBean scheme from a long ways back for naming and implementation.  This should not be a requirement any longer.  This object naming is here to map those old calls.

connection-api/src/main/java/org/terracotta/connection/DiagnosticsConfig.java
myronkscott: Here is the usage pattern 
```
try (Diagnostics d = DiagnosticsFactory.create(<inetaddr>, props)) {
   System.out.println(d.threadDump());
}
```
versus having to create the connection and pull the entity reference.  When the Diagnostics entity is closed the underlying connection will be closed as well.

common-spi/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.terracotta.internal</groupId>
+    <artifactId>terracotta-apis-root</artifactId>
+    <version>1.7-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.terracotta</groupId>
+  <artifactId>common-spi</artifactId>
```
ramsai1729, line 32: I would keep this module in terracotta-core as this used for extending the platform impl.

myronkscott: interesting thought.  i'll look into it.

PR #316, opened 2020-05-28
add hostname to the server environment

myronkscott:


GaryWKeim:
OK, but while you're in there.... Server.getConfiguration is returning the startup config and not the DC one, so changes made through config-tool are not reflected.

PR #317, opened 2020-06-04
Add support for dynamic topologies

ramsai1729:


myronkscott:
should we make a clean break and deprecate the URI and static List methods now?  They make little sense given the way they should be used now.

myronkscott:
anything that uses a URI will be a wrapper above this API.

chrisdennis (approved):
Not sure I understand the context here... but the code doesn't look broken.

PR #321, opened 2020-08-03
support audit logging on the server

myronkscott:
Provide a way for any component to write messages to the audit logs of the server.

### DIFF
--- a/standard-cluster-services/src/main/java/org/terracotta/server/Server.java
+++ b/standard-cluster-services/src/main/java/org/terracotta/server/Server.java
@@ -183,8 +183,22 @@ public interface Server {
    * @return
    */
   Properties getCurrentChannelProperties();
-
+  /**
+   * log messages to the console of the server
+   * @param message message to send
+   * @param sub additional parameters
+   */
   void console(String message, Object...sub);
-
+  /**
+   * log messages to the warning of the server
+   * @param warning message to send
+   * @param event additional parameters
+   */
   void warn(String warning, Object...event);
+  /**
+   * log messages to the auditing of the server
+   * @param message message to send
+   * @param event additional parameters
+   */
+  void audit(String message, Properties additionalIdInfo);
 }


myronkscott:
@saurabhagas or @mathieucarbou Is audit logging going to be used at bootstrap or only runtime?

saurabhagas:
> @saurabhagas or @mathieucarbou Is audit logging going to be used at bootstrap or only runtime?

Both - there's some audit logging during server startup (e.g. parsing whitelist file), and other during runtime (e.g. connections from clients)

standard-cluster-services/src/main/java/org/terracotta/server/Server.java
```diff
@@ -183,8 +183,22 @@
...
+   * @param warning message to send
+   * @param event additional parameters
+   */
   void warn(String warning, Object...event);
+  /**
+   * log messages to the auditing of the server
+   * @param message message to send
+   * @param event additional parameters
+   */
+  void audit(String message, Properties additionalIdInfo);
```
saurabhagas, line 203: This interface is painfully overloaded - it has several disjoint responsibilities. While we're here, would it be a good time to split this interface into multiple short, meaningful interfaces?

myronkscott: we are simply talking about auditing here.  not validation.

saurabhagas, line 203: Curious to know how the usage of this will look like. Would a raise a PR in tc-core to demonstrate the usage?

saurabhagas: Server states, CLI process arguments, logging, performing actions (e.g. dump) etc.

myronkscott: Ah.  this interface itself.  No I don't think we want to go there.  There are several implementors and consumers.

saurabhagas: Okay. Yeah, was talking about `Server` interface, not the new method.

myronkscott: this roughly translates to `Guardian.validate(AUDIT_OP,additionalIdInfo)` with some additions

myronkscott:
> > @saurabhagas or @mathieucarbou Is audit logging going to be used at bootstrap or only runtime?
> 
> Both - there's some audit logging during server startup (e.g. parsing whitelist file), and other during runtime (e.g. connections from clients)

this is problematic in the current implementation.  This service is bootstrapped itself very late so is not available during configuration.

mathieucarbou:
> > > @saurabhagas or @mathieucarbou Is audit logging going to be used at bootstrap or only runtime?
> > 
> > 
> > Both - there's some audit logging during server startup (e.g. parsing whitelist file), and other during runtime (e.g. connections from clients)
> 
> this is problematic in the current implementation. This service is bootstrapped itself very late so is not available during configuration.

@saurabhagas : the way we have implemented it by decorating the DC service and listening, the audit logs methods are only called when the server has reached a final state (like STARTING, ACTIVE, etc).

saurabhagas:
@myronkscott what are the next steps for this PR? Do we merge so that that tc-core changes can be done based on the new API?

PR #330, opened 2022-06-10
Remove snapshots

cljohnso:
This commit stream includes:

* Recover BasicConnectTest and update connection-api documentation

  This commit

  * recovers the BasicConnectTest class (used for documentation samples)
  * updates Asciidoctor plug-in (dropping need for rubygems:coderay)
  * make adjustments to documentation to update for changes over time

* Remove POM repositories elements

  This commit reduces the repositories and pluginRepositories elements
  to those required to locate the direct dependencies of this project.
  More specifically, the SNAPSHOT repositories are removed -- inclusion
  of SNAPSHOT repositories interferes with the use of range-based version
  specifications.  Should a developer want to include a SNAPSHOT release
  in a build, local addition of the SNAPSHOT repositories can be employed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,12 +98,9 @@
 
   <repositories>
     <repository>
-      <id>terracotta-snapshots</id>
-      <url>http://www.terracotta.org/download/reflector/snapshots</url>
-    </repository>
-    <repository>
       <id>terracotta-releases</id>
-      <url>http://www.terracotta.org/download/reflector/releases</url>
+      <url>https://repo.terracotta.org/maven2</url>
+      <snapshots><enabled>false</enabled></snapshots>
     </repository>
   </repositories>
 
--- a/connection-api/pom.xml
+++ b/connection-api/pom.xml
@@ -31,32 +31,11 @@
   <artifactId>connection-api</artifactId>
 
   <properties>
-    <asciidoctor.version>1.5.2</asciidoctor.version>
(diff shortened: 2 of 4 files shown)


GaryWKeim (approved):

pom.xml
```diff
@@ -97,13 +97,10 @@
   </distributionManagement>
 
   <repositories>
-    <repository>
-      <id>terracotta-snapshots</id>
-      <url>http://www.terracotta.org/download/reflector/snapshots</url>
-    </repository>
     <repository>
       <id>terracotta-releases</id>
       <url>http://www.terracotta.org/download/reflector/releases</url>
```
akomakom, line 102: It may make sense to take this opportunity to change http://www.terracotta.org/download/reflector/releases to https://repo.terracotta.org/maven2

cljohnso: @akomakom  If that's that it should be, certainly.  Should it be changed _universally_ ... in all projects I touch?

cljohnso: Updated

akomakom (approved):

chrisdennis (approved):

pom.xml
akomakom: > @akomakom If that's that it should be, certainly. Should it be changed _universally_ ... in all projects I touch?

I would say yes, as the opportunity arises.

PR #332, opened 2024-07-26
Switch to IBM copyright

myronkscott:


connection-api/src/assembly/docs.xml
```diff
@@ -1,20 +1,18 @@
...
-  ~  Software distributed under the License is distributed on an "AS IS" basis,
-  ~  WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
-  ~  the specific language governing rights and limitations under the License.
-  ~
-  ~  The Covered Software is Connection API.
-  ~
-  ~  The Initial Developer of the Covered Software is
-  ~  Terracotta, Inc., a Software AG company
-  ~
+Copyright Terracotta, Inc.
```
cljohnso, line 2: The `license-maven-plugin` is somewhat strict about the whitespace needed in and around a copyright/license block.  Its rules for `xml` files are particularly restrictive.  Use `license:format` to make adjustments to this file so it will pass muster.  (You might ask "why doesn't it complain about the `pom.xml` files?"  Because the inherited configuration explicitly ignores them.)

cljohnso (changes requested):

cljohnso (approved):

